PR #28, opened 2024-08-12
Config Syncer

xvi10:


contracts/config/ConfigSyncer.sol
```diff
@@ -0,0 +1,173 @@
...
+
+    // @dev validate that the market within data is equal to market
+    // @param baseKey the base key to validate
+    // @param market the market address
+    // @param data the data used to compute fullKey
+    function _validateMarketInData(bytes32 baseKey, address market, bytes memory data) internal pure {
+        address marketFromData;
+        if (baseKey == Keys.MAX_PNL_FACTOR) {
+            bytes32 unusedBytes32;
+            bool unusedBool;
```
xvi10: i don't think these are needed, could just comment out the unused variables, example: https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/token/TokenUtils.sol#L62

SentinelDev1: I've updated per the example provided in the latest commit. I also changed the local variable naming from unusedBytes32 and unusedBool to extKey and isLong respectively, for more informative naming. 

Variable name 'extKey' stands for extension or extended key, which is the naming I created and used in utils/configSyncer.ts for the additional key used for parameters with baseKey = MAX_PNL_FACTOR as these were unnamed in the updateMarketConfigUtils.ts file I derived them from and I wanted to differentiate them from baseKey to avoid confusion.

If preferred, I can revert this naming change.

SentinelDev1: I think this is resolved, but I will let you confirm on your end in case I am somehow missing something here.

PR #32, opened 2024-08-21
update gmx market caps

gmxer:


config/markets.ts
```diff
@@ -1372,11 +1372,11 @@ const config: {
...
 
-      maxLongTokenPoolAmount: expandDecimals(107_000, 18),
-      maxShortTokenPoolAmount: expandDecimals(3_000_000, 6),
+      maxLongTokenPoolAmount: expandDecimals(160_000, 18),
+      maxShortTokenPoolAmount: expandDecimals(4_000_000, 6),
 
-      maxLongTokenPoolUsdForDeposit: decimalToFloat(2_500_000),
-      maxShortTokenPoolUsdForDeposit: decimalToFloat(2_500_000),
+      maxLongTokenPoolUsdForDeposit: decimalToFloat(4_000_000),
+      maxShortTokenPoolUsdForDeposit: decimalToFloat(4_000_000),
```
gdev8317: i think it should be $3,5m (for both long and short)

we usually use a bit lower values for caps "ForDeposit"

this is to allow for swaps even if it's not possible to deposit funds into the pool

gdev8317 (approved):

PR #36, opened 2024-09-03
config-syncer updates

SentinelDev1:
Created new PR given the old one was inadvertently closed and merged.


scripts/updateRiskOracleConfig.ts
```diff
@@ -1,51 +1,85 @@
...
+      console.info(`updating: ${syncConfigKey.description}`);
+    }
   }
 
-  console.info(`updating ${multicallWriteParams.length} features`);
+  console.info(`updating for ${multicallWriteParams.length} keys`);
   console.info("multicallWriteParams", multicallWriteParams);
 
-  if (process.env.WRITE) {
+  if (multicallWriteParams.length > 0 && process.env.WRITE === "true") {
```
gdev8317: can it be similar to updateMarketConfigUtils?

https://github.com/gmx-io/gmx-synthetics/blob/updates/scripts/updateMarketConfigUtils.ts#L555

there is only one check `if (write) ...`

scripts/updateRiskOracleConfig.ts
```diff
@@ -1,51 +1,85 @@
...
-    //   data: encodeData(["address"], ["0x352f684ab9e97a6321a13CF03A61316B681D9fD2"]),
-    // },
-    {
+  if (!riskOracleConfig) {
+    throw new Error("No configuration found for the current network");
+  }
+
+  const syncConfigKeys = [];
+  
+  for (const marketAddress in riskOracleConfig.markets) {
```
gdev8317: you can use `for (const [marketAddress, marketConfig] of Object.entries(riskOracleConfig.markets) { ... }`

it's better to not use `for in` to avoid unexpected behavior (for example, with arrays it will return both elements and some array properties)

config/riskOracle.ts
```diff
@@ -1,22 +1,74 @@
...
-      riskOracleConfig: "0x48b67764dBB6B8fc2A0c3987ed3819e543212Bc3", // Not yet deployed to avalanche, using Arbitrum Sepolia address as a placeholder
+      riskOracle: "0x48b67764dBB6B8fc2A0c3987ed3819e543212Bc3", // Not yet deployed to avalanche, using Arbitrum Sepolia address as a placeholder
     },
     avalancheFuji: {
-      riskOracleConfig: "0xE05354F4187820bF0832bF1f5fAd6a0F592b8fB6",
+      riskOracle: "0xE05354F4187820bF0832bF1f5fAd6a0f592b8fB6",
+      markets: {
+        "0xD996ff47A1F763E1e55415BC4437c59292D1F415": { // Address for current Avalanche Fuji deployment AVAX/USDC pool
+          syncConfigMarketDisabled: false,
+          marketParameters: {
```
gdev8317: wondering maybe can be just "parameters"? it's already inside "markets", so it should be clear that it belongs to markets?

scripts/updateRiskOracleConfig.ts
```diff
@@ -1,51 +1,85 @@
...
+    const marketConfig = riskOracleConfig.markets[marketAddress];
+    syncConfigKeys.push({
```
gdev8317: could you check how we handle these in e.g. updateMarketConfig https://github.com/gmx-io/gmx-synthetics/blob/updates/scripts/updateMarketConfigUtils.ts#L507

it first reads all values from DataStore https://github.com/gmx-io/gmx-synthetics/blob/updates/scripts/updateMarketConfigUtils.ts#L528

then it updates params if only it doesn't match values in DataStore https://github.com/gmx-io/gmx-synthetics/blob/updates/scripts/updateMarketConfigUtils.ts#L548

without this current script will always send transaction

scripts/updateRiskOracleConfig.ts
SentinelDev1: Sure, the additional checks were mainly helpful for debugging but I will remove them and better align with the provided example

scripts/updateRiskOracleConfig.ts
SentinelDev1: Thanks for the tip, that makes sense, I will update accordingly.

config/riskOracle.ts
SentinelDev1: Ok, sure I can update, I was just afraid it might be confusing given the ability to disable parameters for all markets as well.

scripts/updateRiskOracleConfig.ts
SentinelDev1: Ok, but I already included a check for this in updateRiskOracleConfig.ts: 
https://github.com/gmx-io/gmx-synthetics/blob/config-syncer/scripts/updateRiskOracleConfig.ts#L52-L65 

This is similar to how it was done in updateOracleConfig.ts: 
https://github.com/gmx-io/gmx-synthetics/blob/config-syncer/scripts/updateOracleConfig.ts#L19-L23

Maybe I'm missing something, but isn't this already addressed with the highlighted code? Or maybe the issue is that I'm doing this check after retrieving all the markets, market parameters and parameters from riskOracle.ts?

PR #53, opened 2024-10-01
Buyback and distribute

SentinelDev1:


contracts/error/Errors.sol
```diff
@@ -140,6 +140,8 @@ library Errors {
 
     // FeeHandler errors
     error InvalidClaimFeesInput(uint256 marketsLength, uint256 tokensLength);
+    error InvalidSwapTokenInput(address swapToken, uint256 arrayPosition);
```
xvi10: can we call the token "buybackToken" and call the error "InvalidBuybackTokenInput"? i think that would be clearer to differentiate it from the regular GMX V2 swaps

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,229 @@
...
 
     // @dev claim fees from the specified markets
     // @param markets the markets to claim fees from
-    // @param tokens the fee tokens to claim
+    // @param feeTokens the fee tokens to claim for (GMX/WNT)
+    // @param swapTokens the provided swap token (GMX/WNT)
     function claimFees(
-        address[] memory markets,
-        address[] memory tokens
+        address[] calldata markets,
```
xvi10: i think we should change this to claimFees for only one market and one buybackToken, the logic seems a bit complicated for an array now

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,229 @@
...
+
+        if (markets.length != swapTokens.length) {
+            revert Errors.InvalidClaimFeesInput(markets.length, swapTokens.length);
         }
 
-        address receiver = dataStore.getAddress(Keys.FEE_RECEIVER);
+        address receiver = address(this);
+
+        (uint256 gmxTokenOraclePrice, uint256 gmxTokenOracleDecimals, uint256 gmxTokenDecimals) = _getTokenPricingValues(gmx);
+        (uint256 wntTokenOraclePrice, uint256 wntTokenOracleDecimals, uint256 wntTokenDecimals) = _getTokenPricingValues(wnt);
```
xvi10: do we need to specify "gmx" and "wnt"? i had the impression that the user would just specify the buybackToken, and all the values would be fetched based on the buybackToken, without explicitly stating the token is gmx or wnt, if there is a buyback batchsize and buyback factor defined for that token, that would mean the token can be bought back, is that possible?

contracts/error/Errors.sol
SentinelDev1: I've updated accordingly in the latest commit

contracts/fee/FeeHandler.sol
SentinelDev1: Updated to remove the loop in the latest commit

contracts/fee/FeeHandler.sol
SentinelDev1: Updated so that gmx and wnt are only referenced in the _setAvailableFeeAmounts() private function. 

I'm not sure that it can be avoided in this case given that the buyback factor is defined for gmx and for wnt it is inferred. 

Also, in the claimFees() function the buyback token is no longer specified now that buybackFees() is it's own function, thus when computing the updated available fee amounts for gmx and wnt the tokens need to be explicitly referenced.

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
         dataStore = _dataStore;
         eventEmitter = _eventEmitter;
+        vaultV1 = _vaultV1;
+        gmx = _gmx;
+    }
+
+    // @dev withdraw fees from this contract
+    // @param token the token for which to claim fees
+    // @param amount the amount of fees to claim
+    function withdrawFees(address token, uint256 amount) external nonReentrant onlyFeeKeeper {
```
xvi10: would be good to validate that there is a buybackBatchAmount for the "token", to prevent accidental withdrawal of fee tokens that are not GMX and not WNT

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+    // @param market the markets to claim fees from
+    // @param feeToken the fee tokens to claim
+    function claimFees(address market, address feeToken) external nonReentrant {
+        uint256 version;
+        uint256 buybackGmxFactor;
+        uint256 feeAmount;
+        if (market == address(0)) {
+            version = 1;
+            buybackGmxFactor = _getBuybackGmxFactor(version);
+            feeAmount = IVaultV1(vaultV1).withdrawFees(feeToken, address(this));
```
xvi10: for V1, only Vault.gov can withdraw fees, Vault.gov is a Timelock which also has a function withdrawFees, so this should be `IVaultGovV1(vaultV1.gov()).withdrawFees(feeToken, address(this))` instead

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+    }
 
-        address receiver = dataStore.getAddress(Keys.FEE_RECEIVER);
+    // @dev receive an amount in feeToken by depositing the batchSize amount of the buybackToken
+    // @param feeToken the token to receive with the fee amount calculated via an oracle price
+    // @param buybackToken the token to deposit in the amount of batchSize in return for fees
+    function buybackFees(address feeToken, address buybackToken) external nonReentrant {
+        uint256 batchSize = dataStore.getUint(Keys.buybackBatchAmountKey(buybackToken));
+        if (batchSize == 0) {
+            revert Errors.InvalidBuybackTokenInput(buybackToken);
```
xvi10: this error can be "InvalidBuybackToken" instead of "InvalidBuybackTokenInput"

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+        uint256 buybackGmxFactor = dataStore.getUint(Keys.buybackGmxFactorKey(version));
+        return buybackGmxFactor;
+    }
+
+    function _getMinFeeTokenAmount(
+        address feeToken,
+        address buybackToken,
+        uint256 batchSize
+    ) private view returns (uint256) {
+        (
```
xvi10: should use https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/oracle/ChainlinkPriceFeedUtils.sol#L17 instead

the returned price would already be adjusted based on the token decimals

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+
+        uint256 feeTokenAmount = (batchSize * feeTokenPriceInBuybackToken) / 1e30;
+        return feeTokenAmount;
+    }
+
+    function _getTokenPricingValues(address token) private view returns (uint256, uint256, uint256) {
+        address tokenPriceFeedAddress = dataStore.getAddress(Keys.priceFeedKey(token));
+        uint256 tokenOraclePrice = IFeedAddress(tokenPriceFeedAddress).latestAnswer();
+        uint256 tokenOracleDecimals = IFeedAddress(tokenPriceFeedAddress).decimals();
+        uint256 tokenDecimals = IFeedAddress(token).decimals(); // IERC20().decimals() is not included in the openzeppelin IERC20 interface so using IFeedAddress().decimals() instead but can change if necessary
```
xvi10: there should not be a need to read token decimals if the prices are already adjusted: https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/oracle/ChainlinkPriceFeedUtils.sol#L62

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+    // @dev receive an amount in feeToken by depositing the batchSize amount of the buybackToken
+    // @param feeToken the token to receive with the fee amount calculated via an oracle price
+    // @param buybackToken the token to deposit in the amount of batchSize in return for fees
+    function buybackFees(address feeToken, address buybackToken) external nonReentrant {
+        uint256 batchSize = dataStore.getUint(Keys.buybackBatchAmountKey(buybackToken));
+        if (batchSize == 0) {
+            revert Errors.InvalidBuybackTokenInput(buybackToken);
+        }
+
+        uint256 minFeeTokenAmount = _getMinFeeTokenAmount(feeToken, buybackToken, batchSize);
```
xvi10: could call this expectedFeeTokenAmount

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+        uint256 maxFeeTokenAmount = (minFeeTokenAmount * (maxPriceImpactFactor + 10000)) / 10000;
+
+        uint256 availableFeeAmount = _getAvailableFeeAmount(feeToken, buybackToken);
+
+        if (availableFeeAmount >= maxFeeTokenAmount) {
+            _buybackFees(feeToken, buybackToken, batchSize, maxFeeTokenAmount, availableFeeAmount);
+        } else if (availableFeeAmount >= minFeeTokenAmount) {
+            _buybackFees(feeToken, buybackToken, batchSize, availableFeeAmount, availableFeeAmount);
+        } else {
+            revert Errors.InsufficientFeeAmount(feeToken, buybackToken, availableFeeAmount);
```
xvi10: i don't think this is necessary, maybe let the user pass in a "minOutputAmount" and validate that the amount that would be sent to the user is >= minOutputAmount, error should be InsufficientBuybackOutputAmount to separate it from the regular fee and swap errors

contracts/fee/IFeedAddress.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: BUSL-1.1
```
xvi10: this should not be needed, the Chainlink logic is already in https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/oracle/ChainlinkPriceFeedUtils.sol#L62

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+            feeAmount = FeeUtils.claimFees(dataStore, eventEmitter, market, feeToken, address(this));
+            _setAvailableFeeAmounts(version, feeToken, feeAmount);
         }
+    }
 
-        address receiver = dataStore.getAddress(Keys.FEE_RECEIVER);
+    // @dev receive an amount in feeToken by depositing the batchSize amount of the buybackToken
+    // @param feeToken the token to receive with the fee amount calculated via an oracle price
+    // @param buybackToken the token to deposit in the amount of batchSize in return for fees
+    function buybackFees(address feeToken, address buybackToken) external nonReentrant {
```
xvi10: let's call this buyback instead of buybackFees

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+        uint256 batchSize = dataStore.getUint(Keys.buybackBatchAmountKey(buybackToken));
```
xvi10: we should also validate that feeToken and buybackToken are not the same, error can be BuybackAndFeeTokenAreEqual

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
-    function claimFees(
-        address[] memory markets,
-        address[] memory tokens
-    ) external nonReentrant onlyFeeKeeper {
-        if (markets.length != tokens.length) {
-            revert Errors.InvalidClaimFeesInput(markets.length, tokens.length);
+    // @param market the markets to claim fees from
+    // @param feeToken the fee tokens to claim
+    function claimFees(address market, address feeToken) external nonReentrant {
+        uint256 version;
```
xvi10: should validate that there is no buybackBatchAmount for the feeToken, GMX and WNT fees should not go through this claim and buyback process

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
         eventEmitter = _eventEmitter;
+        vaultV1 = _vaultV1;
+        gmx = _gmx;
+    }
+
+    // @dev withdraw fees from this contract
+    // @param token the token for which to claim fees
+    // @param amount the amount of fees to claim
+    function withdrawFees(address token, uint256 amount) external nonReentrant onlyFeeKeeper {
+        address receiver = dataStore.getAddress(Keys.FEE_RECEIVER);
```
xvi10, line 57: should allow "markets[]" to be passed in and call FeeUtils.claimFees(dataStore, eventEmitter, market, token, address(this)), to allow all WNT or GMX to be directly claimed from the markets

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,176 @@
...
+        IERC20(token).safeTransfer(receiver, amount);
```
xvi10: could change this to `IERC20(token).safeTransfer(receiver, IERC20(token).balanceOf(address(this));` so that amount does not need to be passed in

contracts/fee/FeeHandler.sol
SentinelDev1: No problem, I’ve added that validation. I had been thinking the same, but wasn't sure if we wanted the ability to withdraw tokens that weren't GMX or WNT in case of an emergency or for some other reason.

contracts/fee/FeeHandler.sol
SentinelDev1: I've updated for this but I'm still assigning vaultV1 as an address in storage instead of as an contract/interface as is implied in the above example because vaultV1 needs to be passed as an address in the IVaultGovV1().withdrawFees() function.

contracts/fee/FeeHandler.sol
SentinelDev1: No problem, I've updated accordingly.

contracts/fee/FeeHandler.sol
SentinelDev1: Thanks, I've updated to use this instead.

contracts/fee/FeeHandler.sol
SentinelDev1: As per the above comment, this is no longer needed and has been removed.

contracts/fee/FeeHandler.sol
SentinelDev1: Updated to expectedFeeTokenAmount.

contracts/fee/FeeHandler.sol
SentinelDev1: Updated to add minOutputAmount as an input param and updated the error to InsufficientBuybackOutputAmount. The error InsufficientBuybackOutputAmount doesn’t include minOutputAmount as a param because it is passed by the user to the function so they should already have that information, but it can be added as well if preferred.

contracts/fee/IFeedAddress.sol
SentinelDev1: As per the above comment, this is no longer needed and has been removed.

contracts/fee/FeeHandler.sol
SentinelDev1: This has now been updated to buyback() from buybackFees().

contracts/fee/FeeHandler.sol
SentinelDev1: This validation has now been added.

contracts/fee/FeeHandler.sol
SentinelDev1: This validation has now been added.

contracts/fee/FeeHandler.sol
SentinelDev1: withdrawFees has now been updated to include an address[] calldata markets input param. Currently, the buybackToken fees are first consolidated in FeeHandler and then transferred to FEE_RECEIVER. If preferred, the claimed fees could be transferred directly to FEE_RECEIVER from the GM market. 

Also, this could be updated to also include an array of buybackTokens to allow for both GMX and WNT to be withdrawn in one call, though I think this would require either matching the length of markets and buybackTokens or a nested for loop.

contracts/fee/FeeHandler.sol
SentinelDev1: This has been updated accordingly. I had been thinking of taking this approach as well, but included the amount input param in case for some reason it would be useful to specify it explicitly.

contracts/fee/FeeHandler.sol
xvi10: the current logic looks ok to me

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,208 @@
...
+    function buyback(address feeToken, address buybackToken, uint256 minOutputAmount) external nonReentrant {
+        if (feeToken == buybackToken) {
+            revert Errors.BuybackAndFeeTokenAreEqual(feeToken, buybackToken);
+        }
+
+        uint256 batchSize = dataStore.getUint(Keys.buybackBatchAmountKey(buybackToken));
+        _validateBuybackToken(batchSize, buybackToken);
+
+        uint256 availableFeeAmount = _getAvailableFeeAmount(feeToken, buybackToken);
+        uint256 maxFeeTokenAmount = _getMaxFeeTokenAmount(feeToken, buybackToken, batchSize);
```
xvi10, line 104: i think this would be clearer as:
```
uint256 outputAmount = availableFeeAmount < maxFeeTokenAmount ? availableFeeAmount : maxFeeTokenAmount

if (buybackAmount < minOutputAmount) { revert ... }

_buybackFees(feeToken, buybackToken, batchSize, outputAmount, availableFeeAmount);
```

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,208 @@
...
+        address buybackToken,
+        uint256 batchSize,
+        uint256 buybackAmount,
+        uint256 availableFeeAmount
+    ) private {
+        IERC20(feeToken).safeTransfer(msg.sender, buybackAmount);
+        IERC20(buybackToken).safeTransferFrom(msg.sender, address(this), batchSize);
+
+        availableFeeAmount = availableFeeAmount - buybackAmount;
+        _setAvailableFeeAmount(feeToken, buybackToken, availableFeeAmount);
```
xvi10: may be slightly clearer as:

 _setAvailableFeeAmount(feeToken, buybackToken, availableFeeAmount - buybackAmount);

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,208 @@
...
+    }
+
+    function _setAvailableFeeAmounts(uint256 version, address feeToken, uint256 feeAmount) private {
```
xvi10: would be clearer to name this _incrementAvailableFeeAmounts to differentiate it from _setAvailableFeeAmount

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,208 @@
...
+    // @param minOutputAmount the minimum amount of the feeToken that the caller will receive
+    function buyback(address feeToken, address buybackToken, uint256 minOutputAmount) external nonReentrant {
+        if (feeToken == buybackToken) {
+            revert Errors.BuybackAndFeeTokenAreEqual(feeToken, buybackToken);
+        }
+
+        uint256 batchSize = dataStore.getUint(Keys.buybackBatchAmountKey(buybackToken));
+        _validateBuybackToken(batchSize, buybackToken);
+
+        uint256 availableFeeAmount = _getAvailableFeeAmount(feeToken, buybackToken);
```
xvi10, line 99: may be good to revert if availableFeeAmount is zero

contracts/fee/FeeHandler.sol
xvi10: would be good to add minOutputAmount to the error for convenience, in the Errors we usually always return (..., "the actual value", "the expected value")

SentinelDev1: Ok no problem, this has been added.

contracts/fee/FeeHandler.sol
SentinelDev1: That makes sense, I've updated per the above suggestion.

contracts/fee/FeeHandler.sol
SentinelDev1: I agree, I’ve updated per your suggestion above.

contracts/fee/FeeHandler.sol
SentinelDev1: No problem, I’ve renamed the function for more clarity.

contracts/fee/FeeHandler.sol
SentinelDev1: Ok, no problem. I added another AvailableFeeAmountIsZero error that returns (address feeToken, address buybackToken, uint256 availableFeeAmount) since it didn't seem to make sense to return minOutputAmount as is now done in the InsufficientBuybackOutputAmount error.

contracts/fee/FeeHandler.sol
SentinelDev1: I don't want to jump the gun, but I think this can be marked resolved now?

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,215 @@
...
+        _buybackFees(feeToken, buybackToken, batchSize, outputAmount, availableFeeAmount);
+    }
+
+    function getOutputAmount(
+        address[] calldata markets,
+        address feeToken,
+        address buybackToken
+    ) external view returns (uint256) {
+        uint256 batchSize = _getBatchSize(buybackToken);
+        _validateBuybackToken(batchSize, buybackToken);
```
gdev8317, line 125: small suggestion, i think we usually don't use prefix "_" for private/internal methods, probably makes sense not to use it here for consistency?

xvi10: we do use prefix "_" for private / internal methods: https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/exchange/OrderHandler.sol#L352

i think only in libraries do we not do that

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,215 @@
...
+
+        uint256 availableFeeAmountGmx = _getAvailableFeeAmount(feeToken, gmx) + feeAmounts.gmx;
+        uint256 availableFeeAmountWnt = _getAvailableFeeAmount(feeToken, wnt) + feeAmounts.wnt;
+
+        _setAvailableFeeAmount(feeToken, gmx, availableFeeAmountGmx);
+        _setAvailableFeeAmount(feeToken, wnt, availableFeeAmountWnt);
+    }
+
+    function _setAvailableFeeAmount(address feeToken, address buybackToken, uint256 availableFeeAmount) private {
+        dataStore.setUint(Keys.buybackAvailableFeeAmountKey(feeToken, buybackToken), availableFeeAmount);
```
gdev8317, line 215: also wondering if it's necessary to create helpers that only call setUint/getUint, at first glance calling dataStore directly is clear/concise as well?

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,215 @@
...
 
         address receiver = dataStore.getAddress(Keys.FEE_RECEIVER);
+        IERC20(buybackToken).safeTransfer(receiver, IERC20(buybackToken).balanceOf(address(this)));
+    }
+
+    // @dev claim fees in feeToken from the specified markets
+    // @param market the market from which to claim fees
+    // @param feeToken the fee tokens to claim from the market
+    function claimFees(address market, address feeToken) external nonReentrant {
+        if (_getBatchSize(feeToken) != 0) {
```
gdev8317: just checking should it be `== 0` instead?

gdev8317: like here https://github.com/gmx-io/gmx-synthetics/pull/53/files#diff-537965269a61a84567326928bad5ef66fa8b7a71cff643c237907d9dba0e7e17R213

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,215 @@
...
+            revert Errors.InvalidFeeToken(feeToken);
+        }
+
+        uint256 feeAmount;
+        if (market == address(0)) {
```
gdev8317: i would suggest for `claimFees` func to accept `uint version` arg to make it more explicit and not rely on market

gdev8317: and maybe create constants like `const uint V1 = 1` and use them instead of plain values 1 and 2

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,215 @@
...
+
+    function _getMaxFeeTokenAmount(
+        address feeToken,
+        address buybackToken,
+        uint256 batchSize
+    ) private view returns (uint256) {
+        uint256 feeTokenPrice = _getPriceFeedPrice(feeToken);
+        uint256 buybackTokenPrice = _getPriceFeedPrice(buybackToken);
+
+        uint256 expectedFeeTokenAmount = Precision.mulDiv(batchSize, feeTokenPrice, buybackTokenPrice);
```
gdev8317: do i understand correctly, that buybackToken is GMX? in this case i guess it should be
```
Precision.mulDiv(batchSize, buybackTokenPrice, feeTokenPrice);
```

otherwise if batch size 100, GMX token price is $20 and fee token = USDC then expected fee token amount = 100 * 1 / 20 = 5 USDC. and i guess it should be 2000 USDC?

contracts/fee/FeeHandler.sol
xvi10: i think this is fine, we do it in multiple other places as well: https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/market/MarketUtils.sol#L1986

contracts/fee/FeeHandler.sol
xvi10: "!= 0" is correct, we do not want the user to be able to claim fees for the buyback tokens (GMX, WNT)

could add a comment to make it clearer

xvi10: "!= 0" is correct, we do not want the user to be able to claim fees for the buyback tokens (GMX, WNT)

could add a comment to make it clearer

contracts/fee/FeeHandler.sol
xvi10: ok with me

contracts/fee/FeeHandler.sol
xvi10: i think you're right, this should be changed

contracts/fee/FeeHandler.sol
SentinelDev1: Ok, I've updated per G's recommendation above in my latest commit, including adding v1 and v2 const uint256 variables

contracts/fee/FeeHandler.sol
SentinelDev1: Yup, this was also identified in the audit as one of the findings and has been corrected in the latest commit

contracts/error/Errors.sol
```diff
@@ -139,7 +139,13 @@ library Errors {
...
 
     // FeeHandler errors
-    error InvalidClaimFeesInput(uint256 marketsLength, uint256 tokensLength);
+    error InvalidFeeToken(address feeToken);
+    error InvalidBuybackToken(address buybackToken);
+    error InvalidVersion(uint256 version);
+    error InsufficientBuybackOutputAmount(address feeToken, address buybackToken, uint256 outputAmount, uint256 minOutputAmount);
+    error BuybackAndFeeTokenAreEqual(address feeToken, address buybackToken);
+    error AvailableFeeAmountIsZero(address feeToken, address buybackToken, uint256 availableFeeAmount);
+    error PricesOlderThanMaxPriceAge(uint256 priceTimestamp, uint256 buybackMaxPriceAge, uint256 currentTimestamp);
```
xvi10: Could be named MaxBuybackPriceAgeExceeded

contracts/data/Keys.sol
```diff
@@ -1998,4 +2000,10 @@ library Keys {
             token
         ));
     }
+
+    // @dev key for the buyback max price age
+    // @return key for buyback max price age
+    function buybackMaxPriceAgeKey() internal pure returns (bytes32) {
```
xvi10: if there is no params required for the key, then usually we just refer to it directly, so a function is not needed, e.g.: https://github.com/gmx-io/gmx-synthetics/blob/main/contracts/oracle/Oracle.sol#L244

contracts/fee/FeeHandler.sol
```diff
@@ -60,26 +61,33 @@ contract FeeHandler is ReentrancyGuard, RoleModule, BasicMulticall {
...
         }
 
         uint256 feeAmount;
-        if (market == address(0)) {
+        if (version == v1) {
             feeAmount = IVaultGovV1(vaultV1.gov()).withdrawFees(address(vaultV1), feeToken, address(this));
-            _incrementAvailableFeeAmounts(1, feeToken, feeAmount);
-        } else {
+        } else if (version == v2) {
             feeAmount = FeeUtils.claimFees(dataStore, eventEmitter, market, feeToken, address(this));
```
xvi10, line 74: should validate that market is not address(0) here

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,221 @@
...
+
+        uint256 availableFeeAmountGmx = _getAvailableFeeAmount(feeToken, gmx) + feeAmounts.gmx;
+        uint256 availableFeeAmountWnt = _getAvailableFeeAmount(feeToken, wnt) + feeAmounts.wnt;
+
+        _setAvailableFeeAmount(feeToken, gmx, availableFeeAmountGmx);
+        _setAvailableFeeAmount(feeToken, wnt, availableFeeAmountWnt);
+    }
+
+    function _setAvailableFeeAmount(address feeToken, address buybackToken, uint256 availableFeeAmount) private {
+        dataStore.setUint(Keys.buybackAvailableFeeAmountKey(feeToken, buybackToken), availableFeeAmount);
```
xvi10, line 215: should emit a SetAvailableFeeAmount event here

contracts/fee/FeeHandler.sol
```diff
@@ -3,46 +3,221 @@
...
+        }
+
+        uint256 maxFeeTokenAmount = _getMaxFeeTokenAmount(feeToken, buybackToken, batchSize);
+        uint256 outputAmount = availableFeeAmount < maxFeeTokenAmount ? availableFeeAmount : maxFeeTokenAmount;
+
+        if (outputAmount < minOutputAmount) {
+            revert Errors.InsufficientBuybackOutputAmount(feeToken, buybackToken, outputAmount, minOutputAmount);
+        }
 
+        _buybackFees(feeToken, buybackToken, batchSize, outputAmount, availableFeeAmount);
```
xvi10, line 111: should emit a Buyback event here

contracts/fee/FeeHandler.sol
gdev8317: i see, i had an impression that this prefix is an exception and rarely used

contracts/fee/FeeHandler.sol
gdev8317: maybe it's fine if these methods are later used in some other contracts, but in this case it only used inside this contract

it might be a matter of taste, i would personally prefer not to add more helpers, in this specific case they don't make code clearer in my view, but increase the code source size / adds some complexity to understand that code because i need to do an extra step to check what's inside `_setAvailableFeeAmount`

xvi10: discussed that this can be separately decided on

contracts/fee/FeeHandler.sol
SentinelDev1: Ok, this has been added in the latest commit

contracts/fee/FeeHandler.sol
SentinelDev1: Ok, this has been added in the latest commit, not sure if buybackToken (and maybe feeToken) should be indexed?

contracts/fee/FeeHandler.sol
SentinelDev1: Ok, this has been added in the latest commit, not sure if buybackToken (and maybe feeToken) should be indexed?

contracts/data/Keys.sol
SentinelDev1: Ok, this has been removed in the latest commit and the base key is now referenced directly in FeeHandler.sol

contracts/error/Errors.sol
SentinelDev1: Ok, this has been renamed in the latest commit

contracts/fee/FeeHandler.sol
xvi10: i think it is fine without index

PR #54, opened 2024-10-01
Eigen token

gmxer:
didn't add "Max PnL Factor Long", as it seems to be same as in base config
`dataStreamFeedId` and `dataStreamFeedDecimals` is pending 
`priceFeed.address` is not 100% approved to be correct

all parameters that equal to default values elliminated in this commit 77b3bd911a3c644781c3c88084ac109cacad51d5

### DIFF
--- a/config/markets.ts
+++ b/config/markets.ts
@@ -866,6 +866,66 @@ const config: {
       thresholdForStableFunding: decimalToFloat(5, 2), // 5%
       thresholdForDecreaseFunding: decimalToFloat(0), // 0%
     },
+
+    {
+      tokens: { indexToken: "EIGEN", longToken: "WETH", shortToken: "USDC" },
+      virtualTokenIdForIndexToken: hashString("PERP:DOGE/USD"),
+      virtualMarketId: hashString("SPOT:DOGE/USD"),
+
+      ...synthethicMarketConfig,
+
+      positionImpactExponentFactor: decimalToFloat(22, 1), // 2.2
+      positivePositionImpactFactor: decimalToFloat(4, 9), // 4e-9,
+      negativePositionImpactFactor: decimalToFloat(8, 9), // 8e-9
+
+      positiveMaxPositionImpactFactor: percentageToFloat("0.5%"), // default
+      negativeMaxPositionImpactFactor: percentageToFloat("0.5%"), // default
+      maxPositionImpactFactorForLiquidations: bigNumberify(0), // default
+      swapImpactExponentFactor: decimalToFloat(2, 0), // default
+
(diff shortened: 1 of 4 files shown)


config/oracle.ts
```diff
@@ -158,6 +158,13 @@ export default async function (hre: HardhatRuntimeEnvironment): Promise<OracleCo
             heartbeatDuration: (24 + 1) * 60 * 60,
           },
         },
+        EIGEN: {
+          priceFeed: {
+            address: "0xD3631AC9D81eD560D61957a55E9c992cdE497eb6",
```
xvi10: may not be needed, as i am unsure if there will be an on-chain feed for EIGEN

would be good to set the placeholder address to "FIXME", to prevent accidentally sending an incorrect configuration

config/markets.ts
```diff
@@ -868,8 +868,8 @@ const config: {
     },
     {
       tokens: { indexToken: "EIGEN", longToken: "WETH", shortToken: "USDC" },
-      virtualTokenIdForIndexToken: hashString("PERP:DOGE/USD"),
-      virtualMarketId: hashString("SPOT:DOGE/USD"),
+      virtualTokenIdForIndexToken: hashString("PERP:EIGEN/USD"),
+      virtualMarketId: hashString("SPOT:WETH/USD"),
```
gmxer, line 863: 🙏🏻

config/tokens.ts
```diff
@@ -121,6 +121,13 @@ const config: {
       dataStreamFeedDecimals: 18,
       oracleTimestampAdjustment: 1,
     },
+    EIGEN: {
+      synthetic: true,
```
xvi10: i don't think we should use synthetic for this, since the token is available on Arbitrum

https://x.com/CamelotDEX/status/1841088733560320393

https://arbiscan.io/address/0x606c3e5075e5555e79aa15f1e9facb776f96c248

config/markets.ts
```diff
@@ -866,6 +866,50 @@ const config: {
       thresholdForStableFunding: decimalToFloat(5, 2), // 5%
       thresholdForDecreaseFunding: decimalToFloat(0), // 0%
     },
+    {
+      tokens: { indexToken: "EIGEN", longToken: "WETH", shortToken: "USDC" },
+      virtualTokenIdForIndexToken: hashString("PERP:EIGEN/USD"),
+      virtualMarketId: hashString("SPOT:WETH/USD"),
+
+      ...synthethicMarketConfig,
+
```
xvi10, line 866: need to add:
```
      maxLongTokenPoolAmount,
      maxShortTokenPoolAmount,
      maxLongTokenPoolUsdForDeposit,
      maxShortTokenPoolUsdForDeposit
```

PR #55, opened 2024-10-01
check if market supported by risk oracle when updating params

gmxer:


scripts/updateMarketConfigUtils.ts
```diff
@@ -685,3 +695,34 @@ export async function updateMarketConfig({ write }) {
     console.log();
   }
 }
+
+async function getSupportedByRiskOracleMarkets(markets, tokens, onchainMarketsByTokens) {
+  const address =
+    hre.network.name === "avalancheFuji"
```
xvi10: should get the address from config/riskOracle instead?

scripts/updateMarketConfigUtils.ts
```diff
@@ -1,33 +1,39 @@
...
   onchainMarketsByTokens,
+  supportedByRiskOracleMarkets,
   tokens,
   generalConfig,
   handleConfig: handleConfigArg,
 }) => {
-  const shouldHandleBaseKey = (baseKey: string) => {
-    if (hre.network.name !== "arbitrum") {
+  const shouldHandleBaseKey = (baseKey: string, isSupportedByRiskOracle: boolean) => {
+    if (hre.network.name !== "arbitrum" && hre.network.name !== "avalancheFuji") {
```
xvi10: may be clearer to add a const SUPPORTED_MARKETS above instead, similar to RISK_ORACLE_MANAGED_BASE_KEYS

scripts/updateMarketConfigUtils.ts
```diff
@@ -597,10 +603,13 @@ export async function updateMarketConfig({ write }) {
   const configKeys = [];
   const multicallReadParams = [];
 
+  const supportedByRiskOracleMarkets = await getSupportedByRiskOracleMarkets(markets, tokens, onchainMarketsByTokens);
```
xvi10: may be a bit easier to read if this is called "supportedRiskOracleMarkets"

PR #59, opened 2024-10-07
Add code to remove updated at block values

xvi10:
When updated contracts from https://github.com/gmx-io/gmx-synthetics/pull/50/files are deployed, the previous contracts would still be functional, these contracts would continue to write block values to the DataStore

These block values should continue to be removed in the updated contracts to avoid dangling values in the DataStore


xvi10:
Closing as this data would be fixed using a separate contract instead if needed

PR #62, opened 2024-10-07
Add APE and SATS configs

xvi10:


config/markets.ts
```diff
@@ -1063,6 +1057,52 @@ const config: {
...
+      thresholdForDecreaseFunding: decimalToFloat(0), // 0%
+
+      minCollateralFactor: decimalToFloat(1, 2), // 1%
+
+      minCollateralFactorForOpenInterestMultiplierLong: decimalToFloat(25, 10), // 2.5e-9
+      minCollateralFactorForOpenInterestMultiplierShort: decimalToFloat(25, 10), // 2.5e-9
+
+      maxOpenInterestForLongs: decimalToFloat(500_000),
+      maxOpenInterestForShorts: decimalToFloat(500_000),
+
```
gdev8317, line 1090: we use increased reserve factor 100% for synthetic markets with borrowing kink model, just checking if it's expected to have default value for STX (SATS?) and APE instead

config/markets.ts
```diff
@@ -1063,6 +1057,52 @@ const config: {
       maxLongTokenPoolAmount: expandDecimals(25, 8), // ~1,5M USD (x3 of max open interest)
       maxShortTokenPoolAmount: expandDecimals(1_500_000, 6), // ~1,5M USD (x3 of max open interest)
     },
+    {
+      tokens: { indexToken: "STX", longToken: "WBTC.e", shortToken: "USDC" },
```
gdev8317: should it be SATS instead of STX?

config/markets.ts
xvi10: will check with Chaos Labs

PR #66, opened 2024-10-23
SUI, SEI, APE

dmax10:


config/markets.ts
```diff
@@ -1603,6 +1603,94 @@ const config: {
...
+      aboveOptimalUsageBorrowingFactor: percentageToFloat("160%").div(SECONDS_PER_YEAR),
+
+      positionImpactPoolDistributionRate: bigNumberify(0),
+      minPositionImpactPoolAmount: bigNumberify(0),
+
+      maxOpenInterest: decimalToFloat(1_000_000),
+      maxPoolUsdForDeposit: decimalToFloat(1_500_000), // 1.5x the max open interest
+
+      maxLongTokenPoolAmount: expandDecimals(763, 18), // ~2M USD (2x the max open interest)
+      maxShortTokenPoolAmount: expandDecimals(1_036_000, 6), // ~2M USD (2x the max open interest)
```
xvi10: for simplicity, could assume that USDC is 1 USD, also the value seems incorrect as it is not 2M USD, it should be `expandDecimals(2_000_000, 6)` instead

config/markets.ts
```diff
@@ -1603,6 +1603,94 @@ const config: {
...
+      aboveOptimalUsageBorrowingFactor: percentageToFloat("160%").div(SECONDS_PER_YEAR),
+
+      positionImpactPoolDistributionRate: bigNumberify(0),
+      minPositionImpactPoolAmount: bigNumberify(0),
+
+      maxOpenInterest: decimalToFloat(500_000),
+      maxPoolUsdForDeposit: decimalToFloat(1_000_000),
+
+      maxLongTokenPoolAmount: expandDecimals(763, 18), // ~2M USD (2x the max open interest)
+      maxShortTokenPoolAmount: expandDecimals(4_655_000, 6), // ~2M USD (2x the max open interest)
```
xvi10: seems incorrect as it is not 2M USD, it should be `expandDecimals(2_000_000, 6)` instead

gmxer (approved):

PR #81, opened 2024-11-20
Add configs for BOME, MEME, FLOKI, MEW

dmax10:


SentinelDev1 (approved):

config/markets.ts
```diff
@@ -1718,6 +1718,138 @@ const config: {
...
+      ...borrowingRateConfig_HighMax_WithHigherBase,
+
+      negativePositionImpactFactor: exponentToFloat("5e-10"),
+      positivePositionImpactFactor: exponentToFloat("2.5e-10"),
+      positionImpactExponentFactor: exponentToFloat("2.2e0"),
+
+      negativeSwapImpactFactor: exponentToFloat("5e-9"),
+      positiveSwapImpactFactor: exponentToFloat("2.5e-9"),
+
+      minCollateralFactor: percentageToFloat("0.02%"), // 50x leverage
```
gmxer: should be 2%, otherwise it's 1 / 0.0002 = 5000x leverage

gmxer: and same for MEME

gmxer (changes requested):

config/markets.ts
dmax10: updated

gmxer (approved):
there's a pending question regarding minCollateralFactor for BOME/MEME but it's still good to be listed that way, we will clarify that later

PR #87, opened 2024-12-04
V2.2

xvi10:


contracts/deposit/DepositUtils.sol
```diff
@@ -38,6 +38,15 @@ library DepositUtils {
     // @param executionFee the execution fee for keepers
     // @param callbackGasLimit the gas limit for the callbackContract
     struct CreateDepositParams {
+        CreateDepositParamsAdresses addresses;
```
gdev8317: there is typo, should be Addresses instead of Adresses

dmax10: fixed

contracts/multichain/LayerZeroProvider.sol
```diff
@@ -0,0 +1,526 @@
...
+        bytes calldata /*extraData*/
+    ) external payable {
+        MultichainUtils.validateMultichainProvider(dataStore, from);
+        MultichainUtils.validateMultichainEndpoint(dataStore, msg.sender);
+
+        /// @dev The `account` field is user-supplied and not validated; any address may be provided by the sender
+        // msg.sender and `from` need to be validated for `amountLD` and `token` to not be spoofable
+        // additionally, the lzReceive, _credit and _debit functions of the OFT should be checked for any overrides
+        // before the `amountLD` and `token` values can be trusted
+        (address account, uint256 srcChainId, uint256 amountLD, bytes memory data) = _decodeLzComposeMsg(message);
```
shankars99, line 105: in the event that `_decodeLzComposeMsg(...args)` fails (due to a malformed `composeMsg`) the users tokens would be stuck in the composer forever. 

having a `try...catch` on this and a failed guid state is a solution since you know:
1. sender on source 
2. srcEid
3. amount

but this can realistically only happen if the front end and composer are not aligned on `composeMsg` OR if someone builds their own FE and makes a mistake

contracts/multichain/LayerZeroProvider.sol
```diff
@@ -0,0 +1,526 @@
...
+            srcChainId
+        );
+
+        // note that for these functions, the remaining gas must be sufficient
+        // otherwise the function will revert
+        // if the action cannot be completed due to gas issues, the user would
+        // need to manually call the function with the right amount of gas
+        // srcChainId could be zero when returned from _decodeLzComposeMsg, skip follow up actions in this case
+        if (srcChainId != 0 && data.length != 0) {
+            (ActionType actionType, bytes memory actionData) = abi.decode(data, (ActionType, bytes));
```
shankars99, line 134: another instance where `data` can be malformed and therefore not decode which _can_ result in a permanent lock of tokens

PR #88, opened 2024-12-10
Add FIL configs

dmax10:


scripts/validateMarketConfigsUtils.ts
```diff
@@ -310,6 +310,11 @@ const recommendedMarketConfig = {
       expectedSwapImpactRatio: 20_000,
       expectedPositionImpactRatio: 20_000,
     },
+    FIL: {
+      negativePositionImpactFactor: exponentToFloat("5e-10"),
```
gmxer, line 314: also needs
```
negativeSwapImpactFactor: exponentToFloat("3.5e-9"),
```

gmxer (changes requested):

scripts/validateMarketConfigsUtils.ts
dmax10: in scripts/validateMarketConfigsUtils.ts
for synthetic markets , the negativeSwapImpactFactor is not needed because there is no swap of e.g. FIL for USDC, because FIL is synthetic

Script seems to run without errors, but please double check.

gmxer (approved):

PR #91, opened 2025-01-10
Gasless orders support

gdev8317:
Added contracts that work with Gelato Relay and pay relay fee:
- BaseGelatoRelayRouter: abstract contract with base functionality:
  - validate message's signature
  - validate message's nonce and deadline
  - handle ERC20 token permits
  - handle relay fee (swap to required token, pay relay fee)
  - create/updated/cancel
- GelatoRelayRouter
  - a router contract, accepts calls signed by a user
- SubaccountGelatoRelayRouter
  - a router contract, accepts calls signed by a whitelisted subaccount and creates order for the corresponding main account

Flow:
- each function accepts relay params:
  - relay fee params: token and amount that user is paying for relay fee and execution fee. relay fee is paid to fee collector. residual fee token amount is used for execution fee
  - oracle params: used to swap user's fee token to relay fee token, can be empty if swap is not needed
  - token permits: to update allowance in ERC20 contracts
  - external calls: could be used to swap relay fees through external exchange instead of v2 pools (for gas optimisation). works only with GelatoRelayRouter, i.e. subaccounts don't support external calls
- each function requires erc712 signature 
- custom nonce is used to prevent replay attack
- optional deadline, can limit signature's expiration time
- SubaccountGelatoRelayRouter functions also accept subaccount approval: an instruction signed by account (not subaccount) to whitelist subaccount, update max allowed actions or subaccount expiration time 
- after all validations OrderHandler is used to create, update or cancel order

Other changes:
- added subaccountExpiresAt, it is possible to allow subaccount to trade on behalf of account up until deadline instead up until max allowed actions. both actions count and deadline are validated
- changed SubaccountUtils functions to external to optimize router contracts size
- updated Oracle._setPrices to allow empty prices (prices can be empty if there is no need to swap fee token)
- GLV_MAX_MARKET_TOKEN_BALANCE_USD and GLV_MAX_MARKET_TOKEN_BALANCE_AMOUNT are added to allowed limited base keys
- for orders with a callback contract created by subacounts the execution fee is validated against the max execution fee to prevent the subaccount from stealing account funds. Subaccounts should validate that the execution fee is not too high. excessive execution fee is sent to holding address

Reasons for using non ERC2771 version:
- we will be able to display human readable payload in users wallet when signing transaction data
- ERC2771 doesn't allow sending transactions to a blockchain different from selected in a user's wallet. we'll need this feature for multi-chain and it probably makes sense to use the same logic for all routers
- no need to trust a forwarder

Deployment notes:
- contracts: GelatoRelayRouter, SubaccountGelatoRelayRouter, SubaccountRouter (there were some changes), OrderHandler, Config, Oracle, ExchangeRouter (existing one should also work fine)
- roles for new routers: CONTROLLER and ROUTER_PLUGIN
- configure `maxExecutionFeeMultiplierFactor`


contracts/error/Errors.sol
```diff
@@ -420,4 +423,13 @@ library Errors {
 
     // Reader errors
     error EmptyMarketPrice(address market);
+
+    // Gelato relay errors
+    error InvalidFeeToken(address feeToken, address expectedFeeToken);
+    error NotSupported();
```
xvi10: this doesn't seem to be used

contracts/error/Errors.sol
```diff
@@ -420,4 +423,13 @@ library Errors {
 
     // Reader errors
     error EmptyMarketPrice(address market);
+
+    // Gelato relay errors
+    error InvalidFeeToken(address feeToken, address expectedFeeToken);
```
xvi10: i think InvalidRelayFeeToken might be clearer

contracts/router/BaseGelatoRelayRouter.sol
```diff
@@ -0,0 +1,337 @@
...
+        address account,
+        bytes32 orderKey,
+        address uiFeeReceiver,
+        address residualFeeReceiver
+    ) internal returns (uint256) {
+        _handleTokenPermits(tokenPermit);
+        return _handleRelayFee(contracts, fee, account, orderKey, uiFeeReceiver, residualFeeReceiver);
+    }
+
+    function _handleTokenPermits(TokenPermit[] calldata tokenPermit) internal {
```
xvi10: i think this should be called tokenPermits instead of tokenPermit

contracts/router/SubaccountGelatoRelayRouter.sol
```diff
@@ -0,0 +1,162 @@
...
+        address subaccount;
+        uint256 expiresAt;
+        uint256 maxAllowedCount;
+        bytes32 actionType;
+        uint256 deadline;
+        uint256 nonce; // for replay attack protection
+        bytes signature;
+    }
+
+    bytes32 public constant _SUBACCOUNT_APPROVAL_TYPEHASH =
```
xvi10: just curious, why is there an underscore for this?

gdev8317: no specific reason, have removed

contracts/oracle/Oracle.sol
```diff
@@ -179,7 +179,11 @@ contract Oracle is RoleModule {
     // @param params OracleUtils.SetPricesParams
     function _setPrices(
         OracleUtils.ValidatedPrice[] memory prices
-    ) internal returns (OracleUtils.ValidatedPrice[] memory) {
+    ) internal {
+        if (prices.length == 0) {
```
xvi10, line 184: when will prices.length be zero? it may be useful to add a comment here with an example of this case

gdev8317: added comment

> // in case of gasless relay the prices are not required if there is no need to swap fee tokens

PR #93, opened 2025-01-12
Use balanceWasImproved instead of priceImpactUsd > 0

dmax10:


contracts/pricing/PositionPricingUtils.sol
```diff
@@ -197,24 +197,30 @@ library PositionPricingUtils {
...
-                impactExponentFactor
+            return (
+                PricingUtils.getPriceImpactUsdForCrossoverRebalance(
+                    initialDiffUsd,
+                    nextDiffUsd,
+                    positiveImpactFactor,
+                    negativeImpactFactor,
+                    impactExponentFactor
+                ),
+                false // balanceWasImproved
```
xvi10: should be "balanceWasImproved" instead of "false"

dmax10: Updated

contracts/reader/ReaderDepositUtils.sol
```diff
@@ -35,6 +35,11 @@ library ReaderDepositUtils {
         ISwapPricingUtils.SwapPricingType swapPricingType;
     }
 
+    struct PriceImpactCache {
```
xvi10: should name this GetDepositAmountOutCache

dmax10: Renamed.

PR #94, opened 2025-01-13
Fee Automation

xvi10:


contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
 
     // Reader errors
     error EmptyMarketPrice(address market);
+
+    // MultichainReader errors
+    error NotEnoughNative(uint256 msgValue);
```
xvi10: rename to InsufficientMultichainNativeFee

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error LzTokenUnavailable();
```
xvi10: where is this used?

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error OnlyEndpoint(address addr);
```
xvi10: should use Errors.Unauthorized instead

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error ConfirmationsLengthMismatch(uint256 chainIdsLength, uint256 confirmationsLength);
```
xvi10: where is this used?

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error UnauthorizedOriginator(address originator);
```
xvi10: should use Errors.Unauthorized instead

contracts/external/MultichainReader.sol
```diff
@@ -0,0 +1,230 @@
...
+            received,
+            readData
+        );
+        receivedData.received = true;
+        receivedData.readData = _message;
+
+        IOriginator(originator).processLzReceive(_guid, receivedData);
+        bytes memory transactionCallData = IOriginator(originator).transactionCallData(_guid);
+        if (transactionCallData.length != 0) {
+            (bool success, ) = originator.call(transactionCallData);
```
xvi10: this doesn't seem safe, MultichainReader will have the CONTROLLER role, it should not call generic functions

`IOriginator(originator).processLzReceive(_guid, receivedData);` already seems sufficient, why is `originator.call(transactionCallData);` needed?

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error OriginatorCallFailed(bytes transactionCallData);
+    error OnlyPeer(uint32 eid, bytes32 sender);
```
xvi10: use Errors.Unauthorized instead

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error NoPeer(uint32 eid);
```
xvi10: rename to EmptyPeer

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error InvalidEndpointCall();
```
xvi10: where is this used?

contracts/error/Errors.sol
```diff
@@ -420,4 +420,16 @@ library Errors {
...
+    error InvalidDelegate();
```
xvi10: where is this used?

contracts/error/Errors.sol
SentinelDev1: Ok, no problem I can rename it as suggested. Just to note, most of these errors including this one were copied from the LayerZero code that was previously imported/inherited, but I don't see why we can't change the name on our end.

contracts/error/Errors.sol
SentinelDev1: This was used in the _payLzToken() function that was previously inherited from the OAppSender.sol contract. After refactoring to remove the various imports I removed because to minimize complexity, the MultichainReader contract is designed to pay using only the native token so it didn't seem necessary. If you agree using only native tokens to pay is sufficient, I will go ahead and remove this redundant error.

contracts/error/Errors.sol
SentinelDev1: This is another error copied from the LayerZero contracts and thus using LayerZero naming, but I can update the naming as requested on our end.

contracts/error/Errors.sol
SentinelDev1: This is an error I added earlier before using dataStore when number of confirmations per chain was being stored directly in the contract (the chain Ids and number of confirmations were array input params in the setter function - it was to make sure there wasn't a length mismatch between the two input arrays). Now it is being stored using the "MULTICHAIN_CONFIRMATIONS" key in dataStore thus is no longer used. I will remove it in the next commit.

contracts/error/Errors.sol
SentinelDev1: This is another error copied from the LayerZero contracts and thus using LayerZero naming, but I can update the naming as requested on our end.

contracts/error/Errors.sol
SentinelDev1: The first is our own error message so I will remove and use Unauthorized instead, the second is another error using LayerZero naming but can also change on our end.

contracts/error/Errors.sol
SentinelDev1: This is another error copied from the LayerZero contracts and thus using LayerZero naming, but I can update the naming as requested on our end.

contracts/error/Errors.sol
SentinelDev1: It was in the IOAppCore interface which is no longer imported, I don't recall exactly where it was used but as it doesn't seem to be used I will remove it.

contracts/error/Errors.sol
SentinelDev1: It was in the IOAppCore interface which is no longer imported, I don't recall exactly where it was used but as it doesn't seem to be used I will remove it.

contracts/error/Errors.sol
SentinelDev1: The first is our own error message, and given the concerns around originator.call(transactionCallData) it may no longer be necessary anyway. The second is another error using LayerZero naming but can also be changed on our end.

contracts/external/MultichainReader.sol
SentinelDev1: Are you suggesting the function in transactionCallData should be internal/private and invoked via processLzReceive in the originator contract? If so, internal functions can't be called with calldata the same way (via .call(calldata)), so the internal function would need to be explicitly defined in processLzReceive. This approach is fine but I think would require defining the function at the time of deploying the originator contract, reducing the flexibility of defining it dynamically during sendReadRequests(). Let me know if you agree with this alternate approach or have something else in mind.

contracts/error/Errors.sol
xvi10: i think using the native token to pay is fine

contracts/external/MultichainReader.sol
```diff
@@ -0,0 +1,230 @@
...
+
+    // TODO: add modifier for access control
+    function setDelegate(address _delegate) external {
+        endpoint.setDelegate(_delegate);
+    }
+
+    function sendReadRequests(
+        MultichainReaderUtils.ReadRequestInputs[] calldata readRequestInputs,
+        MultichainReaderUtils.ExtraOptionsInputs calldata extraOptionsInputs
+    ) external payable returns (MessagingReceipt memory, bytes32, MultichainReaderUtils.ReceivedData memory) {
```
xvi10: might make sense to have an onlyController modifier here

contracts/external/MultichainReader.sol
xvi10: > Are you suggesting the function in transactionCallData should be internal/private and invoked via processLzReceive in the originator contract? 

yes

> If so, internal functions can't be called with calldata the same way (via .call(calldata)), so the internal function would need to be explicitly defined in processLzReceive

i think that is fine, i believe most originator contracts would just have one internal function to call on processLzReceive

if needed, the originator contract can have a mapping, e.g. mapping (bytes32 => ResponseHandler) guidHandlers

where ResponseHandler is an enum, and in processLzReceive the originator contract can determine the correct function to call based on the guid and the corresponding enum

> This approach is fine but I think would require defining the function at the time of deploying the originator contract, reducing the flexibility of defining it dynamically during sendReadRequests

i think we need to balance flexibility and security, even though there may not be an immediate risk to allowing the MultichainReader to call a generic function on the originator, and the originator contract should validate the caller for each of its functions, in case there is some missing access control, a combination of flexibility and development mistakes could lead to a compromise

minimizing access control over flexibility should be preferred where possible / practicle, especially where in this case the workaround of a mapping does not add too much overhead

contracts/external/MultichainReader.sol
SentinelDev1: Sorry, to clarify, is this referring to the setDelegate() function? 

I was thinking of removing it as previously discussed, but perhaps a safer option would be to use a highly restrictive modifier like onlyController. 

LayerZero responded to my questions around this on telegram saying that a delegate could be useful if faulty dvns need to be removed for example, even if other configuration changes aren't necessary.

xvi10: i'm referring to adding an onlyController modifier to sendReadRequests, because in lzReceive, the originator contract is called, and the originator contract would be validating that msg.sender is MultichainReader, so MultichainReader has some unique access to call potentially important contracts

so i think it would be safer to restrict access of who can call MultichainReader

contracts/data/Keys.sol
```diff
@@ -467,6 +467,54 @@ library Keys {
...
+    // @dev key for FeeDistributor chain ID
+    bytes32 public constant FEE_DISTRIBUTOR_CHAIN_ID = keccak256(abi.encode("FEE_DISTRIBUTOR_CHAIN_ID"));
+    // @dev key for FeeDistributor fee amount
+    bytes32 public constant FEE_DISTRIBUTOR_FEE_AMOUNT = keccak256(abi.encode("FEE_DISTRIBUTOR_FEE_AMOUNT"));
+    // @dev key for FeeDistributor total fee amount in USD by version
+    bytes32 public constant FEE_DISTRIBUTOR_FEE_AMOUNT_USD = keccak256(abi.encode("FEE_DISTRIBUTOR_FEE_AMOUNT_USD"));
+    // @dev key for FeeDistributor staked gmx
+    bytes32 public constant FEE_DISTRIBUTOR_STAKED_GMX = keccak256(abi.encode("FEE_DISTRIBUTOR_STAKED_GMX"));
+    // @dev key for FeeDistributor maximum bridging slippage
+    bytes32 public constant FEE_DISTRIBUTOR_MAX_SLIPPAGE = keccak256(abi.encode("FEE_DISTRIBUTOR_MAX_SLIPPAGE"));
```
xvi10: could rename this to FEE_DISTRIBUTOR_MAX_BRIDGE_SLIPPAGE

contracts/data/Keys.sol
```diff
@@ -467,6 +467,54 @@ library Keys {
...
+    // @dev key for FeeDistributor referral rewards for a given timestamp
+    bytes32 public constant FEE_DISTRIBUTOR_REFERRAL_REWARDS_AMOUNT = keccak256(abi.encode("FEE_DISTRIBUTOR_REFERRAL_REWARDS_AMOUNT"));
+    // @dev key for FeeDistributor max read response delay in seconds from MultichainReader
+    bytes32 public constant FEE_DISTRIBUTOR_MAX_READ_RESPONSE_DELAY = keccak256(abi.encode("FEE_DISTRIBUTOR_MAX_READ_RESPONSE_DELAY"));
+    // @dev key for FeeDistributor gas limit used for the MultichainReader read request
+    bytes32 public constant FEE_DISTRIBUTOR_GAS_LIMIT = keccak256(abi.encode("FEE_DISTRIBUTOR_GAS_LIMIT"));
+    // @dev key for FeeDistributor number of chains with GMX deployments
+    bytes32 public constant FEE_DISTRIBUTOR_NUMBER_OF_CHAINS = keccak256(abi.encode("FEE_DISTRIBUTOR_NUMBER_OF_CHAINS"));
+    // @dev key for FeeDistributor chain ID
+    bytes32 public constant FEE_DISTRIBUTOR_CHAIN_ID = keccak256(abi.encode("FEE_DISTRIBUTOR_CHAIN_ID"));
```
xvi10: i think we could use a chainId uint256 array in datastore instead of FEE_DISTRIBUTOR_NUMBER_OF_CHAINS, FEE_DISTRIBUTOR_CHAIN_ID

EDIT: i think you mentioned that you are already working on changing this

contracts/data/Keys.sol
```diff
@@ -467,6 +467,54 @@ library Keys {
...
+    // @dev key for FeeDistributor staked gmx
+    bytes32 public constant FEE_DISTRIBUTOR_STAKED_GMX = keccak256(abi.encode("FEE_DISTRIBUTOR_STAKED_GMX"));
+    // @dev key for FeeDistributor maximum bridging slippage
+    bytes32 public constant FEE_DISTRIBUTOR_MAX_SLIPPAGE = keccak256(abi.encode("FEE_DISTRIBUTOR_MAX_SLIPPAGE"));
+    // @dev key for FeeDistributor read response timestamp
+    bytes32 public constant FEE_DISTRIBUTOR_READ_RESPONSE_TIMESTAMP = keccak256(abi.encode("FEE_DISTRIBUTOR_READ_RESPONSE_TIMESTAMP"));
+    // @dev key for FeeDistributor LayerZero version of chainId
+    bytes32 public constant FEE_DISTRIBUTOR_LAYERZERO_CHAIN_ID = keccak256(abi.encode("FEE_DISTRIBUTOR_LAYERZERO_CHAIN_ID"));
+    // @dev key for contract and keeper addresses used in FeeDistributor
+    bytes32 public constant FEE_DISTRIBUTOR_STORED_ADDRESSES = keccak256(abi.encode("FEE_DISTRIBUTOR_STORED_ADDRESSES"));
```
xvi10: could rename this to FEE_DISTRIBUTOR_ADDRESS_INFO

contracts/fee/FeeDistributor.sol
```diff
@@ -4,199 +4,384 @@ pragma solidity ^0.8.0;
...
+                    IERC20(gmx).balanceOf(feeKeeper);
+                uint256 stakedGmx = IERC20(feeGmxTracker).totalSupply();
+                dataStore.setUint(Keys.feeDistributorFeeAmountKey(chainId), feeAmount);
+                dataStore.setUint(Keys.feeDistributorStakedGmxKey(chainId), stakedGmx);
+                skippedCurrentChain = true;
+                continue;
+            }
+
+            uint32 layerZeroChainId = uint32(dataStore.getUint(Keys.feeDistributorLayerZeroChainIdKey(chainId)));
+            uint256 readRequest = skippedCurrentChain ? (i - 1) * 3 : i * 3;
```
xvi10: could rename this to readRequestIndex

contracts/fee/FeeDistributor.sol
```diff
@@ -4,199 +4,384 @@ pragma solidity ^0.8.0;
...
+            address feeGmxTracker = dataStore.getAddress(
+                Keys.feeDistributorStoredAddressesKey(chainId, keccak256(abi.encode("FEE_GMX_TRACKER")))
+            );
+            if (chainId == block.chainid) {
+                uint256 feeAmount = dataStore.getUint(Keys.withdrawableBuybackTokenAmountKey(gmx)) +
+                    IERC20(gmx).balanceOf(feeKeeper);
+                uint256 stakedGmx = IERC20(feeGmxTracker).totalSupply();
+                dataStore.setUint(Keys.feeDistributorFeeAmountKey(chainId), feeAmount);
+                dataStore.setUint(Keys.feeDistributorStakedGmxKey(chainId), stakedGmx);
+                skippedCurrentChain = true;
```
xvi10: is skippedCurrentChain needed, or can the chainId array exclude the current chain?

contracts/data/Keys.sol
SentinelDev1: No problem, this will be reflected in my next commit.

contracts/data/Keys.sol
SentinelDev1: Yes, this was updated in my most recent commit.

contracts/data/Keys.sol
SentinelDev1: No problem, this will be reflected in my next commit.

contracts/fee/FeeDistributor.sol
SentinelDev1: No problem, this will be reflected in my next commit.

contracts/fee/FeeDistributor.sol
SentinelDev1: I thought quite a bit about this as well, and we could take that approach, but decided against it for 2 reasons:

1. I felt that having the ChainId array consistent for the dataStore on each chain would be more manageable if and when updates are made, as the same data could be passed to each chain without having to take into account removing the current chain when populating or updating.
2. In order for the bridging logic on each chain to deterministically calculate the exact same amount of GMX bridged from one chain to another chain, I think it simplifies things to have the array the same on each chain.

contracts/fee/FeeDistributor.sol
```diff
@@ -4,199 +4,453 @@ pragma solidity ^0.8.0;
...
+            for (uint256 i; i < chainCount; i++) {
+                difference[i] = int256(feeAmount[i]) - int256(target[i]);
+            }
+
+            uint256[][] memory bridging = new uint256[][](chainCount);
+            for (uint256 i; i < chainCount; i++) {
+                bridging[i] = new uint256[](chainCount);
+            }
+
+            uint256 deficit;
```
xvi10: it is a bit difficult to understand the code here, i think that indexes should have the word index in the name to make it clearer, e.g. deficitIndex

also if there can be some comments that would help

i also think that the code could be re-written to be easier to understand, for example:
```
                while (deficit < chainCount && difference[deficit] >= 0) {
                    deficit++;
                }
                if (deficit == chainCount) break;
```
this seems to be moving the index forward until a chain with deficit is found, but the outer for loop is already looping through each chain, so it does not seem like moving the index forward within the loop is necessary

contracts/error/Errors.sol
```diff
@@ -420,4 +420,19 @@ library Errors {
 
     // Reader errors
     error EmptyMarketPrice(address market);
+
+    // MultichainReader errors
+    error InsufficientMultichainNativeFee(uint256 msgValue);
+    error EmptyPeer(uint32 eid);
+
+    // FeeDistributor errors
+    error DistributionThisWeekAlreadyCompleted(uint256 lastDistributionTime, uint256 startOfCurrentWeek);
```
xvi10: could rename this to "FeeDistributionAlreadyCompleted"

contracts/fee/FeeDistributor.sol
```diff
@@ -4,199 +4,467 @@ pragma solidity ^0.8.0;
...
+            }
         }
 
-        return feeBatch;
+        address extendedGmxTracker = dataStore.getAddress(
+            Keys.feeDistributorAddressInfoKey(block.chainid, extendedGmxTrackerKey)
+        );
+        feeDistributorVault.transferOut(gmxCurrentChain, extendedGmxTracker, feeAmountCurrentChain);
+
+        address gmxRewardDistributor = IRewardTracker(extendedGmxTracker).distributor();
```
xvi10: there can be a function:
```
updateRewardDistributionRate(address tracker, amountForWeek) {
  address distributor = IRewardTracker(tracker).distributor();
  IRewardDistributor(distributor).updateLastDistributionTime()
  IRewardDistributor(distributor). setTokensPerInterval(amountForWeek / 1 weeks)
}
```

so that we do not need to remember to call updateLastDistributionTime for each call

SentinelDev1: I've updated this but also changed it a bit so that it takes address rewardToken as an input and instead of having:

`feeDistributorVault.transferOut(gmxCurrentChain, extendedGmxTracker, feeAmountCurrentChain);`

done separately before, it does this transfer as part of the internal function:

```
function updateRewardDistribution(address rewardToken, address tracker, uint256 rewardAmount) internal {
        feeDistributorVault.transferOut(rewardToken, tracker, rewardAmount);
        
        address distributor = IRewardTracker(tracker).distributor();
        IRewardDistributor(distributor).updateLastDistributionTime();
        IRewardDistributor(distributor). setTokensPerInterval(rewardAmount / 1 weeks);
    }
```

I've also updated the name to just updateRewardDistribution since it's also transferring the funds to the tracker.

SentinelDev1: Let me know if you are ok with these changes, prefer keeping the transfer separate, or have any other feedback.

contracts/error/Errors.sol
SentinelDev1: This has been updated.

contracts/fee/FeeDistributor.sol
SentinelDev1: I reviewed the code again and added some comments yesterday to try to help make things a bit clearer and also added "Index" to the indexes as suggested.

contracts/fee/FeeDistributor.sol
```diff
@@ -4,199 +4,522 @@ pragma solidity ^0.8.0;
...
+            setUint(Keys.feeDistributorStakedGmxKey(chainId), stakedGmx);
+        }
+
+        setUint(Keys.FEE_DISTRIBUTION_STATE, uint256(DistributionState.ReadDataReceived));
+
+        EventUtils.EventLogData memory eventData;
+        eventData.boolItems.initItems(1);
+        eventData.boolItems.setItem(0, "distributeReferralRewards", false);
+
+        eventEmitter.emitEventLog("TriggerFeeDistributorKeeper", eventData);
```
xvi10: i think instead of having the same event name TriggerFeeDistributorKeeper, and distributeReferralRewards be true / false

i think there should be different event names

e.g. for here the event name could be "FeeDistributionDataReceived"

in "distribute" the event name can be "FeeDistributionCompleted"

it would also be good for these events to provide more info on the fee distribution, e.g. how much GMX / WNT is distributed in total, how much WNT was sent to GLP, to keepers, treasury, Chainlink

SentinelDev1: I've just pushed an update with events updated as described above, though I named the first event "FeeDistributionReceivedData" instead of "FeeDistributionDataReceived" to better align the naming with other variable names used that reference receivedData

PR #95, opened 2025-01-14
V2.2 small tasks

xvi10:


contracts/adl/AdlUtils.sol
```diff
@@ -93,7 +93,7 @@ library AdlUtils {
         uint256 latestAdlTime = getLatestAdlTime(dataStore, market, isLong);
 
         if (oracle.maxTimestamp() < latestAdlTime) {
-            revert Errors.OracleTimestampsAreSmallerThanRequired(oracle.maxTimestamp(), latestAdlTime);
+            revert Errors.OracleTimestampsAreSmallerThanRequired(market, oracle.maxTimestamp(), latestAdlTime);
```
xvi10: the task is

> Add token address to OracleTimestampsAreSmallerThanRequired error

we want to give the token address not the market address, but i realized we need to store additional information to do this, so will re-check if this task is needed

dmax10: Reverted the commit.

PR #96, opened 2025-01-15
V2.2 small tasks 2

dmax10:


contracts/config/Config.sol
```diff
@@ -210,6 +212,15 @@ contract Config is ReentrancyGuard, RoleModule, BasicMulticall {
     ) external onlyConfigKeeper nonReentrant {
         MarketUtils.distributePositionImpactPool(dataStore, eventEmitter, market);
 
+        // Ensure the full positionImpactPoolAmount cannot be distributed in less then the minimum required time
+        uint256 positionImpactPoolAmount = MarketUtils.getPositionImpactPoolAmount(dataStore, market);
+        uint256 distributionAmount = Precision.applyFactor(MIN_POSITION_IMPACT_POOL_DISTRIBUTION_TIME, positionImpactPoolDistributionRate);
```
xvi10, line 218: this should be MIN_POSITION_IMPACT_POOL_DISTRIBUTION_TIME * positionImpactPoolDistributionRate?

contracts/market/MarketStoreUtils.sol
```diff
@@ -84,6 +84,11 @@ library MarketStoreUtils {
             keccak256(abi.encode(key, SHORT_TOKEN)),
             market.shortToken
         );
+
+        dataStore.setBytes32(
+            keccak256(abi.encode(key, MARKET_SALT)),
```
xvi10: i think this is not needed

contracts/market/MarketStoreUtils.sol
```diff
@@ -111,6 +116,15 @@ library MarketStoreUtils {
         dataStore.removeAddress(
             keccak256(abi.encode(key, SHORT_TOKEN))
         );
+
+        bytes32 salt = dataStore.getBytes32(keccak256(abi.encode(key, MARKET_SALT)));
```
xvi10: the salt should be passed into the remove function as a parameter

test/exchange/MarketIncreaseOrder.ts
```diff
@@ -258,7 +258,7 @@ describe("Exchange.MarketIncreaseOrder", () => {
 
     await handleOrder(fixture, { create: params });
 
-    expect((await provider.getBalance(user1.address)).sub(initialBalance)).closeTo("244862985958904", "10000000000000");
+    expect((await provider.getBalance(user1.address)).sub(initialBalance)).closeTo("129042985032344", "10000000000000");
```
xvi10, line 261: why is there this change? it seems like quite a large change in value

contracts/market/MarketStoreUtils.sol
dmax10: Removed dataStore salt setter and passed salt as param to the remove function

contracts/market/MarketStoreUtils.sol
dmax10: Removed dataStore salt setter and passed salt as param to the remove function

contracts/config/Config.sol
dmax10: positionImpactPoolDistributionRate has FLOAT_PRECISION and we should divide by that precision to have distributionAmount in e18, same as positionImpactPoolAmount (i.e. 1e18 * 1e30 / 1e30).

Added "setPositionImpactDistributionRate reverts if position impact pool is fully distributed in less than 1 week (604800 seconds)" test for this logic change.

xvi10: i see, thank you for the explanation, i think adding a comment here would be good

test/exchange/MarketIncreaseOrder.ts
dmax10: The refunded amount seems so different in percentage terms because the refundFeeAmount is the delta between the executionFee (which is fixed 0.002) and the executionFeeForKeeper (which is increased from 0.00177 to 0.00189)

contracts/config/Config.sol
dmax10: Added comment

xvi10 (approved):

PR #102, opened 2025-01-22
Add a “dataList” field for deposits, orders, withdrawals, shifts etc

dmax10:


contracts/adl/AdlUtils.sol
```diff
@@ -57,6 +57,7 @@ library AdlUtils {
         bool isLong;
         uint256 sizeDeltaUsd;
         uint256 updatedAtTime;
+        bytes32[] dataList;
```
xvi10: should not be needed

contracts/exchange/BaseHandler.sol
```diff
@@ -50,4 +50,11 @@ contract BaseHandler is RoleModule, GlobalReentrancyGuard, OracleModule {
             ErrorUtils.revertWithCustomError(reasonBytes);
         }
     }
+
+    function validateDataSize(uint256 dataLength) internal view {
```
xvi10: rename to validateDataListLength, doesn't seem to be used, should use this function

contracts/exchange/BaseHandler.sol
```diff
@@ -50,4 +50,11 @@ contract BaseHandler is RoleModule, GlobalReentrancyGuard, OracleModule {
...
+        uint256 maxDataLength = dataStore.getUint(Keys.MAX_DATA_LENGTH);
+        if (dataLength > maxDataLength) {
+            revert Errors.MaxDataLengthExceeded(dataLength, maxDataLength);
```
xvi10: rename to MaxDataListLengthExceeded

contracts/glv/glvDeposit/GlvDepositUtils.sol
```diff
@@ -407,7 +409,8 @@ library GlvDepositUtils {
                 executionFee: 0,
                 callbackGasLimit: 0
             }),
-            Deposit.Flags({shouldUnwrapNativeToken: false})
+            Deposit.Flags({shouldUnwrapNativeToken: false}),
+            glvDeposit.dataList()
```
xvi10: dataList is to pass to callbacks, in here, the callbackContract is address(0), so there is no reason to pass glvDeposit.dataList(), should pass an empty list instead

utils/storeUtils.ts
```diff
@@ -157,6 +157,7 @@ export async function validateStoreUtils({
       addresses: {},
       numbers: {},
       flags: {},
+      _dataList: [],
```
xvi10, line 178: should pass in a non-empty list and add validations below that the dataList is properly set, updated and removed

utils/deposit.ts
```diff
@@ -71,6 +71,7 @@ export async function createDeposit(fixture, overrides: any = {}) {
     shouldUnwrapNativeToken,
     executionFee,
     callbackGasLimit,
+    dataList: [],
```
xvi10: should allow dataList to be optionally passed in for deposits, withdrawals, etc

for the existing createDeposit, createWithdrawal, etc, tests, pass in a non-empty dataList and validate that it is properly set after create

contracts/exchange/BaseHandler.sol
dmax10: Renamed

contracts/adl/AdlUtils.sol
dmax10: Removed

utils/deposit.ts
dmax10: Made the dataList optional and tested with non-empty values for create*

utils/storeUtils.ts
dmax10: Added setSampleItemBytes32Array function down bellow which is dynamically setting values for dataList (similar to setting addresses / numbers / flags)

contracts/exchange/BaseHandler.sol
dmax10: Renamed and added validations for deposits, glvDeposits, withdrawals, shifts, orders

contracts/glv/glvDeposit/GlvDepositUtils.sol
dmax10: Changing it to `new bytes32[](0)` slightly  increases the bytecode (by 0.055 KiB) and throws with the contract size exceeded error. 

This would get solved though once the claimable-collateral-reduction-factor branch is merged because it includes some additional ExchangeRouter refactoring. 

Would update this suggestion once that branch is merged.

dmax10: `glvDeposit.dataList()` has been changed to `new bytes32[](0) // dataList`

PR #105, opened 2025-01-27
Multichain

xvi10:


contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,164 @@
...
+        uint256 userNonce,
+        uint256 deadline
+    ) external nonReentrant onlyGelatoRelay returns (bytes32) {
+        bytes32 structHash = _getCreateDepositStructHash(relayParams, params, userNonce, deadline);
+        _validateCall(userNonce, deadline, account, structHash, signature);
+
+        // TODO: confirm virtual account is used correctly
+        address virtualAccount = MultichainUtils.getVirtualAccount(account, sourceChainId);
+
+        return _createDeposit(relayParams.tokenPermits, relayParams.fee, params, virtualAccount);
```
xvi10: should use _createDeposit(..., account, chainId) here instead of virtualAccount

the GM tokens should be sent to MultichainVault and the account's multichain GM balance should be increased

so would need a deposit.chainId as well

in execute deposit, if the deposit.chainId is "0" this means that the deposit would be a normal deposit, so the GM tokens would be sent directly to the user, if the deposit.chainId is not zero, then the user's multichain GM balance should be increased

contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,164 @@
...
+        RelayFeeParams calldata fee,
+        DepositUtils.CreateDepositParams memory params, // can't use calldata because need to modify params.numbers.executionFee
+        address virtualAccount
+    ) internal returns (bytes32) {
+        Contracts memory contracts = Contracts({
+            dataStore: dataStore,
+            eventEmitter: eventEmitter,
+            // TODO: confirm Contracts struct can be modified --> replace `OrderVault orderVault;` field with `StrictBank vault;`
+            // otherwise, should probably overridde _handleRelay
+            orderVault: OrderVault(payable(multichainVault))
```
xvi10: @gdev8317 is it ok for Max to change Contracts { OrderVault orderVault } into Contracts { StrictBank bank }?

contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,164 @@
...
+            tokenPermits,
+            fee,
+            from,
+            NonceUtils.getKey(contracts.dataStore, NonceUtils.getCurrentNonce(dataStore) + 1), // calculate next key without incrementing (createDeposit will recalculate and increment)
+            address(multichainVault) // residualFeeReceiver
+        );
+
+        // TODO: confirm where funds are bridged. Are they sent to this contract, to MultichainProvider, or directly to MultichainVault?
+        // if funds are sent directly to MultichainVault, _sendTokens would not be needed, but then where is the relay fee being paid from?
+        _sendTokens(
```
xvi10: funds should be bridged to MultichainVault directly

> if funds are sent directly to MultichainVault, _sendTokens would not be needed

I don't quite understand this, to create a deposit, the tokens would still need to be sent from MultichainVault to DepositVault?

> then where is the relay fee being paid from?

relay fee should be sent from MultichainVault, from the user's multichain balance

residual fees should be sent to the deposit vault

contracts/multichain/MultichainUtils.sol
```diff
@@ -7,6 +7,6 @@ pragma solidity ^0.8.0;
  */
 library MultichainUtils {
     function getVirtualAccount(address account, uint256 sourceChainId) internal pure returns (address) {
```
xvi10: should remove this function

dmax10: removed

contracts/multichain/MultichainRouter.sol
dmax10: > I don't quite understand this, to create a deposit, the tokens would still need to be sent from MultichainVault to DepositVault?

I thought so. DepositUtils.createDeposit does a recordTransferIn, which assumes tokens have already been transferred to DepositVault

contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,183 @@
...
+import "../router/relay/GelatoRelayRouter.sol";
+import "../deposit/DepositUtils.sol";
+import "../deposit/DepositVault.sol";
+import "../exchange/IDepositHandler.sol";
+
+import "./MultichainVaultHandler.sol";
+import "./MultichainUtils.sol";
+
+contract MultichainRouter is GelatoRelayRouter {
+    struct GaslessCreateDepositParams {
```
gdev8317: i would suggest to call it MulticallCreateDepositParams (and use multichain instead of gasless in otehr names) because GelatoRelayRouter is also gasless

gdev8317: and maybe chainId, longTokenAmount and shortTokenAmount can be passed as a separate arguments and remove the extra nesting? just thinking, at first glance it could simplify the contract interface

contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,183 @@
...
+            params.shortTokenAmount
+        );
+
+        // pay relay fee tokens from MultichainVault to DepositVault and decrement user's multichain balance
+        params.createDepositParams.executionFee = _handleRelay(
+            contracts,
+            tokenPermits,
+            fee,
+            account,
+            NonceUtils.getKey(contracts.dataStore, NonceUtils.getCurrentNonce(dataStore) + 1), // calculate next key without incrementing
```
gdev8317: would it be the same as `NonceUtils.getNextKey(dataStore)`?

contracts/multichain/MultichainUtils.sol
```diff
@@ -2,11 +2,18 @@
...
+import "../data/Keys.sol";
+
 /**
  * @title MultichainUtils
  */
 library MultichainUtils {
-    function getVirtualAccount(address account, uint256 sourceChainId) internal pure returns (address) {
-        return address(uint160(uint256(keccak256(abi.encode("GMX Multichain", account, sourceChainId)))));
+    function increaseBalance(DataStore dataStore, uint256 chainId, address account, address token, uint256 amount) internal {
+        dataStore.decrementUint(Keys.multichainBalanceKey(chainId, account, token), amount);
```
gdev8317: i guess it should be incrementUint instead of decrementUint

contracts/multichain/MultichainRouter.sol
```diff
@@ -0,0 +1,183 @@
...
+    function _createDeposit(
+        TokenPermit[] calldata tokenPermits,
+        RelayFeeParams calldata fee,
+        address account,
+        GaslessCreateDepositParams memory params // can't use calldata because need to modify params.numbers.executionFee
+    ) internal returns (bytes32) {
+        Contracts memory contracts = Contracts({
+            dataStore: dataStore,
+            eventEmitter: eventEmitter,
+            // TODO: confirm Contracts struct can be modified --> replace `OrderVault orderVault;` field with `StrictBank vault;`
```
gdev8317: i think it's okay to change type to StrictBank, yes

contracts/multichain/MultichainEventUtils.sol
```diff
@@ -16,56 +16,56 @@ library MultichainEventUtils {
...
     ) internal {
         EventUtils.EventLogData memory eventData;
 
         eventData.addressItems.initItems(2);
         eventData.addressItems.setItem(0, "token", token);
-        eventData.addressItems.setItem(1, "virtualAccount", virtualAccount);
+        eventData.addressItems.setItem(1, "account", account);
 
         eventData.uintItems.initItems(2);
         eventData.uintItems.setItem(0, "amount", amount);
```
gdev8317, line 52: wondering it if makes sense to only pass deposit key to this event instead of all other fields

events consume quite a lot of gas, unless i'm missing something account/token/amount could be retrieved from corresponding DepositCreated event later

but tbh i'm not sure if the difference worth it. events just consume a lot of gas already unfortunately

contracts/multichain/MultichainUtils.sol
dmax10: Changed it.

contracts/multichain/MultichainRouter.sol
dmax10: Yes, same logic, the difference is that getNextKey also writes the incremented value (it is named "get" but it does "set" as well).

contracts/multichain/MultichainRouter.sol
dmax10: Changed to:
```
struct MultichainCreateDepositParams {
        uint256 desChainId;
        uint256 longTokenAmount;
        uint256 shortTokenAmount;
        DepositUtils.CreateDepositParams createDepositParams;
    }
```
and replaced gasless with multichain as well.

contracts/multichain/MultichainRouter.sol
gdev8317: oh right, not ideal naming, would be good to change it sometime

contracts/multichain/MultichainRouter.sol
dmax10: Changed it.

contracts/multichain/MultichainRouter.sol
dmax10: Updated. Named it srcChainId though.

contracts/multichain/MultichainRouter.sol
dmax10: Didn't change the nesting yet, but could do that as well if needed.

contracts/multichain/MultichainRouter.sol
dmax10: Changed it.

PR #106, opened 2025-01-28
Refactor glv deposit utils to reduce bytecode size

dmax10:
Moved _validateFirstGlvDeposit and _getMintAmount internal methods from GlvDepositUtils to the newly created GlvDepositHelper library.

### DIFF
--- a/contracts/glv/glvDeposit/GlvDepositHelper.sol
+++ b/contracts/glv/glvDeposit/GlvDepositHelper.sol
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.0;
+
+import "./GlvDeposit.sol";
+import "../GlvUtils.sol";
+
+library GlvDepositHelper {
+    using GlvDeposit for GlvDeposit.Props;
+    using SafeCast for int256;
+
+    address public constant RECEIVER_FOR_FIRST_GLV_DEPOSIT = address(1);
+
+    function validateFirstGlvDeposit(
+        DataStore dataStore,
+        GlvDeposit.Props memory glvDeposit
+    ) external view {
+        address glv = glvDeposit.glv();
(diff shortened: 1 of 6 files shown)


contracts/glv/glvDeposit/GlvDepositHelper.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    function getMintAmount(
+        DataStore dataStore,
+        Oracle oracle,
+        GlvDeposit.Props memory glvDeposit,
+        uint256 receivedMarketTokens,
+        uint256 glvValue,
+        uint256 glvSupply
+    ) external view returns (uint256) {
```
gdev8317: wondering is there a reason to create a separate library with external functions instead of making these functions external inside the GlvDepositUtils?

dmax10: The goal of this refactor was to reduce GlvDepositUtils contract size.
So the approach was to create a new lib (i.e. GlvDepositHelper) and move some of the logic from GlvDepositUtils to this new lib. And GlvDepositHelper would de deployed instead of inlined into GlvDepositUtils

gdev8317: right, i've missed that it is used inside the GlvDepositUtils

in this case wondering if we can use a different name for this new lib. "Utils" and "Helper" sounds quite similar

dmax10: we could, but wasn't sure how to name it differently. They are both "utils" so I was kinda looking for a synonym... any naming ideas?

xvi10: can we name it GlvDepositCalc? i don't think it is much better, but i can't think of much other names

dmax10: Renamed to GlvDepositCalc

PR #109, opened 2025-02-03
Added price feeds validation

gmxer:


scripts/initOracleConfigForTokensUtils.ts
```diff
@@ -181,3 +186,49 @@ export async function initOracleConfigForTokens({ write }) {
...
+    return;
+  }
+
+  const contract = new ethers.Contract(
+    priceFeed.address,
+    ["function decimals() view returns (uint8)", "function description() view returns (string)"],
+    ethers.provider
+  );
+  const decimals = await contract.decimals();
+  const description = await contract.description();
```
gdev8317: let's parallelize requests?

scripts/initOracleConfigForTokensUtils.ts
```diff
@@ -181,3 +186,49 @@ export async function initOracleConfigForTokens({ write }) {
...
+  if (decimals !== priceFeed.decimals) {
+    throw new Error(`Decimals mismatch for ${tokenSymbol}: ${decimals} !== ${priceFeed.decimals}`);
+  }
+
+  const tokenSymbolReplaced =
+    {
+      "WBTC.e": "BTC",
+      tBTC: "BTC",
+      WETH: "ETH",
+      "USDC.e": "USDC",
```
gdev8317, line 228: since you decided to add existing tokens, let's also add Avalanche tokens (for example, "BTC.b", "WETH.e" and "USDT.e"), could check in tokens.ts

scripts/initOracleConfigForTokensUtils.ts
```diff
@@ -181,3 +186,49 @@ export async function initOracleConfigForTokens({ write }) {
...
+  if (!priceFeed) {
+    return;
+  }
+
+  const contract = new ethers.Contract(
+    priceFeed.address,
+    ["function decimals() view returns (uint8)", "function description() view returns (string)"],
+    ethers.provider
+  );
+  const decimals = await contract.decimals();
```
gdev8317: it will throw an error if price fee == AddressZero

gdev8317: and also if the call fails let's throw some human readable error so it's easier to understand what happened

gdev8317 (approved):

PR #110, opened 2025-02-06
Contract deploy validation script

Call-Sol:
Task: https://docs.google.com/document/d/1qugdg3t2URwhBrW1veWEX3SHcZWVjsRyT7W9LJKkldM/edit?tab=t.0

Process:

1. Fetch transaction logs for SignalRoleGranted event
2. Check that roles are valid
3. Collect contracts affected
4. For each contract try to validate it via etherscan if it is verified
5. For each contract get its bytecode from locally compiled artifact and compare with eth_getCode() result from blockchain. 
    - get corresponding deployment file from deployments directory and extract contract name and constructor args
    - git checkout and compile contract at audited commit
    - get artifact file
    - get blockchain bytecocde (execute eth_getCode)
    - compare bytecode from artifact and bytecode from blockchain

### DIFF
--- a/scripts/validateContractDeployment.ts
+++ b/scripts/validateContractDeployment.ts
@@ -34,7 +34,7 @@ async function main() {
     const contracts = await validateRoles(tx);
     console.log("Found these contracts with changed roles: ", contracts);
     for (const contract of contracts) {
-      const sourceCodeVerified = await ValidateFromEtherscan(contract);
+      const sourceCodeVerified = await validateFromEtherscan(contract);
       // Fallback to bytecode compilation if sources are not verified on etherscan
       if (!sourceCodeVerified) {
         await compareContractBytecodes(provider, contract);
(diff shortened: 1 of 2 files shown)


scripts/validateContractDeployment.ts
```diff
@@ -0,0 +1,333 @@
...
+  const account = eventArg[4][0][0][0].value;
+  const roleKey = eventArg[4][4][0][0].value;
+  return {
+    account: account,
+    roleKey: roleKey,
+  };
+}
+
+async function checkRole(signal: SignalRoleInfo): Promise<boolean> {
+  const rolesConfig = await roles(hre);
```
xvi10, line 144: i think we should specify the expected roles here instead of using the rolesConfig just for extra safety

so for example, there can be expectedRoles = { CONTROLLER: ["DepositHandler", "OrderHandler", ...], ROLE_ADMIN: ["Timelock"] }

scripts/validateContractDeployment.ts
```diff
@@ -0,0 +1,333 @@
...
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+// Verify sources
+
+async function ValidateFromEtherscan(contractAddress: string): Promise<boolean> {
```
xvi10: this should be `validateFromEtherscan` instead of `ValidateFromEtherscan`?

xvi10 (approved):

scripts/validateContractDeployment.ts
Call-Sol: renamed

scripts/validateContractDeployment.ts
Call-Sol: Expected roles validation added.

scripts/validateContractDeployment.ts
```diff
@@ -0,0 +1,387 @@
...
+}
+
+// Verify sources from etherscan
+async function validateFromEtherscan(contractInfo: ContractInfo): Promise<boolean> {
+  console.log(`Trying to validate ${contractInfo.address} via etherscan`);
+  const apiKey = hre.network.verify.etherscan.apiKey;
+  const url = hre.network.verify.etherscan.apiUrl + "api";
+  try {
+    const path =
+      url + "?module=contract" + "&action=getsourcecode" + `&address=${contractInfo.address}` + `&apikey=${apiKey}`;
```
gdev8317, line 162: jfyi axio accepts options arg and you can do this instead of manually building querystring:
```
axios.get(url, {
  params: {
    module: "contract",
    ...
  }
});
```

Call-Sol: Thanks

PR #113, opened 2025-02-18
refactor callbacks to use orderdata

Call-Sol:
Gas increase is 21k


Call-Sol:
Changed visibility modifiers for better library composition. 

Even if library has an external functions if another contract is using only its internal functions it will be fully included in the sources(i.e. embedded linkage).
In order to split libraries some functions were made external or public

contracts/deposit/DepositEventUtils.sol
```diff
@@ -26,7 +26,7 @@ library DepositEventUtils {
         Deposit.Props memory deposit,
         Deposit.DepositType depositType
     ) external {
-        EventUtils.EventLogData memory eventData;
+        EventUtils.EventLogData memory eventData = createEventData(deposit, depositType);
 
         eventData.addressItems.initItems(6);
```
xvi10: This code and the code below it can be removed?

contracts/callback/CallbackUtils.sol
```diff
@@ -116,14 +123,16 @@ library CallbackUtils {
         bytes32 key,
         Deposit.Props memory deposit,
         EventUtils.EventLogData memory eventData
-    ) internal {
+    ) external {
         if (!isValidCallbackContract(deposit.callbackContract())) { return; }
 
         validateGasLeftForCallback(deposit.callbackGasLimit());
 
+        EventUtils.EventLogData memory depositData = DepositEventUtils.createEventData(deposit, Deposit.DepositType.Normal);
```
xvi10, line 131: i think passing Deposit.DepositType.Normal is not accurate here since this could be a callback for shift or glv?

contracts/callback/CallbackUtils.sol
```diff
@@ -160,14 +171,16 @@ library CallbackUtils {
         bytes32 key,
         Withdrawal.Props memory withdrawal,
         EventUtils.EventLogData memory eventData
-    ) internal {
+    ) external {
         if (!isValidCallbackContract(withdrawal.callbackContract())) { return; }
 
         validateGasLeftForCallback(withdrawal.callbackGasLimit());
 
+        EventUtils.EventLogData memory withdrawalData = WithdrawalEventUtils.createEventData(withdrawal, Withdrawal.WithdrawalType.Normal);
```
xvi10, line 179: i think passing Withdrawal.WithdrawalType.Normal is not accurate here since this could be a callback for shift or glv?

contracts/event/EventUtils.sol
```diff
@@ -21,6 +21,12 @@ library EventUtils {
         StringItems stringItems;
     }
 
+    struct OrderData {
```
xvi10: Where is this used?

contracts/callback/CallbackUtils.sol
Call-Sol: These callbacks are using only by normal deposit/withdrawals. 
Args for callback are normal deposit/withdrawal props as well

contracts/callback/CallbackUtils.sol
Call-Sol: These callbacks are using only by normal deposit/withdrawals. 
Args for callback are normal deposit/withdrawal props as well

xvi10: i see, right

Call-Sol:
Closing this PR and proceeding to https://github.com/gmx-io/gmx-synthetics/pull/115

PR #122, opened 2025-03-03
Add getters to event utils lib

Call-Sol:
Add getters to event utils lib

Add console errors to validate script


contracts/event/EventUtils.sol
```diff
@@ -3,6 +3,8 @@
 pragma solidity ^0.8.0;
 
 library EventUtils {
+    error NotFound(string key);
```
xvi10: this can be moved to Errors.sol and renamed to EventItemNotFound

having all the Errors in one contract makes it easier to parse errors when needed

contracts/error/Errors.sol
```diff
@@ -460,4 +460,7 @@ library Errors {
     error InvalidSubaccountApprovalSubaccount();
     error NonEmptyExternalCallsForSubaccountOrder();
     error InvalidRelayParams();
+
+    // EventUtils error
+    error NotFound(string key);
```
xvi10: should be renamed to EventItemNotFound

PR #127, opened 2025-03-05
GLV updates, JIT

gdev8317:
JIT stands for Just-in-time liquidity. This PR adds support for atomic execution of GLV shift and order in the same transaction. from product perspective the idea is to allow creating orders in markets even if there is not enough liquidity. keeper would detect such orders and shift liquidity to the order's market before executing it

Changes:
- added new JitOrderHandler contract that allow for atomic execution of GLV shifts and orders in the same transaction
- support GLV oracle price. now it can optionally be used to save gas costs in GLV value calculation. GLV oracle price may deviate slightly from calculated GLV price. added note to the `GlvUtils.getGlvValue` function. without this executing GLV shift and an order may not be possible due to high gas usage
- added `simulateExecuteLatestJitOrder` to ExchangeRouter but had to remove existing `simulateExecute*` because of contract's side. for order/deposit/withdrawal creation `simulateExecuteLatest*` functions can be used instead (they don't require `key` argument)
- removed onlyController from simulation functions in *Handler contracts. it didn't make much sense because ExchangeRouter's simulation functions (that were just proxies to *Handlers) were public and permissionless. these functions accept `key` and can be used instead of the ones removed from ExchangeRouter
- renamed GLV_SHIFT_MAX_PRICE_IMPACT_FACTOR -> GLV_SHIFT_MAX_LOSS_FACTOR. in practice it's not exactly price impact and it led to a confusion

Migration notes:
- Keeper should validate shifts are not executed too frequently and not too much funds are spent on them (including gas cost and price impact). As an alternative keeper could limit shifts by volume or spent funds instead of count, this would be more accurate
- Keeper should always use the latest prices for JIT orders
- Keeper should account for JIT_SHIFT_UPDATE_AT_TIME_BUFFER when validating prices
- Keeper should fall back to OrderHandler if JitOrderHandler fails
- GLV shiftMinInterval should be set to 0 after keeper supports shift interval validation. otherwise JIT orders would be blocked by on-chain validation
- UI and all integrations should migrate to new simulateExecuteLatest* functions for simulations
- should run glv config scripts to set GLV_SHIFT_MAX_LOSS_FACTOR 
- keeper should switch to GLV_SHIFT_MAX_LOSS_FACTOR 
- UI should validate max pnl of glvDeposit.market


contracts/exchange/JitOrderHandler.sol
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            (bytes32 glvShiftKey, GlvShift.Props memory glvShift) = _createGlvShift(
+                _dataStore,
+                shiftParamsList[i],
+                order.updatedAtTime(),
+                orderKey
+            );
+
+            glvShiftHandler._executeGlvShift(
```
gdev8317: guess it's better to rename `_executeGlvShift` or create new function, underscore usually means it's private/internal

contracts/exchange/OrderHandler.sol
```diff
@@ -264,11 +263,30 @@ contract OrderHandler is IOrderHandler, BaseOrderHandler {
         }
     }
 
+    // @dev used by other handlers to avoid duplicating the same code on their side
+    // this method is similar to `executeOrder` but skips execution gas validation
+    function executeOrderForController(
```
xvi10: i think this should be named executeOrderFromController, to match e.g. DepositHandler.executeDepositFromController

contracts/deposit/ExecuteDepositUtils.sol
```diff
@@ -71,18 +71,23 @@ library ExecuteDepositUtils {
...
+    // @param skipRemoval if true, the deposit will not be removed from the data store.
+    // This is used when executing a deposit as part of a shift or a glv deposit and the deposit is not stored in the data store
+    function executeDeposit(IExecuteDepositUtils.ExecuteDepositParams memory params, Deposit.Props memory deposit, bool skipRemoval) external returns (uint256 receivedMarketTokens) {
         // 63/64 gas is forwarded to external calls, reduce the startingGas to account for this
         params.startingGas -= gasleft() / 63;
 
-        DepositStoreUtils.remove(params.dataStore, params.key, deposit.account());
-
+        if (!skipRemoval) {
+            DepositStoreUtils.remove(params.dataStore, params.key, deposit.account());
```
xvi10, line 93: wondering if there should be a validation here, e.g.
```
if (dataStore.containsBytes32(Keys.SHIFT_LIST, key)) {
  revert Errors.RemovalShouldNotBeSkipped(Keys.SHIFT_LIST, key);
}
```

just to make it more difficult to incorrectly skip removal when code is updated

PR #129, opened 2025-03-07
min collateral factor for liquidation

Call-Sol:


contracts/position/PositionUtils.sol
```diff
@@ -407,7 +409,15 @@ library PositionUtils {
             + cache.priceImpactUsd
             - collateralCostUsd.toInt256();
 
-        cache.minCollateralFactor = MarketUtils.getMinCollateralFactor(dataStore, market.marketToken);
+        if (forLiquidation) {
+            cache.minCollateralFactor = MarketUtils.getMinCollateralFactorForLiquidation(dataStore, market.marketToken);
+            if (cache.minCollateralFactor == 0) {
```
xvi10: i think this is not needed, the min collateral factor for liquidation should be initialized for all markets https://www.notion.so/gmxio/V2-Deployments-12903574745d80b693d8f1afacebe887

PR #133, opened 2025-03-09
set holding address from timelock

Call-Sol:


deploy/configureGeneralSettings.ts
```diff
@@ -1,13 +1,18 @@
...
+  signalHoldingAddress,
+} from "../utils/dataStore";
 import { updateGeneralConfig } from "../scripts/updateGeneralConfigUtils";
 
 const func = async ({ gmx }: HardhatRuntimeEnvironment) => {
   const generalConfig = await gmx.getGeneral();
 
   await setAddressIfDifferent(keys.FEE_RECEIVER, generalConfig.feeReceiver, "fee receiver");
-  await setAddressIfDifferent(keys.HOLDING_ADDRESS, generalConfig.holdingAddress, "holding address");
+  await signalHoldingAddress(generalConfig.holdingAddress);
```
xvi10: i think this should not be changed, cos we still want to set the holding address directly for testnet

utils/dataStore.ts
```diff
@@ -59,3 +60,9 @@ async function setIfDifferent(
     log("skipping %s %s (%s) as it is already set to %s", type, label, key, value.toString());
   }
 }
+
+export async function signalHoldingAddress(address: string) {
```
xvi10: i think this should be in utils/timelock.ts

deploy/configureGeneralSettings.ts
Call-Sol: Fixed

utils/dataStore.ts
Call-Sol: Removed this function

PR #139, opened 2025-03-12
Gelato sponsored call

gdev8317:
Gelato supports 2 types of calls:
1. callWithSyncFee (relay fee is paid to the Relay within the same transaction)
2. sponsoredCall. Gelato Balance is manually toped up and used to pay for transactions

this PR supports (2). if msg.sender != Gelato then contracts collect relay fee and send it to RELAY_FEE_ADDRESS. this collected fee can later be used to top up Gelato Balance

the flow has changed a bit, the relay fee processing is split into 2 steps:
1. swap fee to wnt if needed and pay execution fee (i.e. keeper for for order execution in separate txn)
2. calculate tnx fee (if sponsored call) and pay the relay fee

step (2) should be called at the end to correctly calculate gas spent during execution

other changes:
- supported skipping signature validation if tx.origin == `address(uint160(uint256(keccak256("GMX SIMULATION ORIGIN"))))`. this will allow estimating gas before signing the message
- added `batch` function to both relay routers to allow for creating, updating and/or cancelling multiple orders
- supported using external calls for order collateral (previously was supported for relay fee only)
- relay fee swap size is limited by MAX_RELAY_FEE_SWAP_USD_FOR_SUBACCOUNT for subaccount orders
- refactored to make EIP-712 typehashes consistent between GelatoRelayRouter and SubaccountGelatoRelayRouter

notes:
- RELAY_FEE_ADDRESS, GELATO_RELAY_FEE_MULTIPLIER_FACTOR, GELATO_RELAY_FEE_BASE_AMOUNT, RELAY_EXECUTION_GAS_FEE_PER_ORACLE_PRICE, MAX_RELAY_FEE_SWAP_USD_FOR_SUBACCOUNT params should be configured. if GELATO_RELAY_FEE_MULTIPLIER_FACTOR the default value 1e30 would be used
- GELATO_RELAY_FEE_BASE_AMOUNT should cover base txn cost, the relay overhead, calldata decoding, gas spent after the relay fee is paid (e.g. residual fee is sent back to the user)
- GELATO_RELAY_FEE_MULTIPLIER_FACTOR should account for the Relay premium
- Gelato Relay should be configured to accept transactions only to whitelisted addresses and methods, otherwise anyone would be able to send transactions and consume Gelato Balance


contracts/gas/GasUtils.sol
```diff
@@ -537,4 +555,73 @@ library GasUtils {
 
         eventEmitter.emitEventLog1("ExecutionFeeRefundCallback", Cast.toBytes32(callbackContract), eventData);
     }
+
+    function payGelatoRelayFee(
```
0xcuriousapple, line 559: @audit test

PR #141, opened 2025-03-18
Allow price impact pool withdrawal

Call-Sol:


contracts/config/ConfigTimelockController.sol
```diff
@@ -12,13 +12,43 @@ import {OracleStore} from "../oracle/OracleStore.sol";
...
+        address[] memory executors,
+        Oracle oracle
+    ) TimelockController(minDelay, proposers, executors, msg.sender) OracleModule(oracle) {}
 
+    function executeAtomicWithOraclePrices(
+        address target,
+        uint256 value,
+        bytes calldata payload,
+        OracleUtils.SetPricesParams calldata oracleParams
+    ) external onlyRoleOrOpenRole(EXECUTOR_ROLE) withOraclePricesForAtomicAction(oracleParams) {
```
xvi10: sorry i think we should use withOraclePrices instead, and validate that oracle.minTimestamp > Chain.currentTimestamp - CONFIG_MAX_PRICE_AGE config

CONFIG_MAX_PRICE_AGE would be a new config

contracts/config/ConfigTimelockController.sol
```diff
@@ -12,13 +12,43 @@ import {OracleStore} from "../oracle/OracleStore.sol";
...
+        execute(target, value, payload, 0, 0);
+    }
+
+    function withdrawFromPositionImpactPool(
+        address market,
+        address receiver,
+        uint256 amount
+    ) external  {
+        require(msg.sender == address(this), "TimelockController: caller must be timelock");
```
xvi10: could use the Errors.Unauthorized error here instead

contracts/error/Errors.sol
```diff
@@ -55,6 +55,7 @@ library Errors {
     error InvalidFeeReceiver(address receiver);
     error InvalidOracleSigner(address signer);
     error InvalidHoldingAddress(address account);
+    error InvalidWithdrawalAmount(uint256 amount);
```
xvi10: could change this to EmptyPositionImpactWithdrawalAmount, i think "amount" param is not needed since it would always be zero

contracts/error/Errors.sol
```diff
@@ -198,6 +199,9 @@ library Errors {
     error InvalidMarketTokenBalanceForClaimableFunding(address market, address token, uint256 balance, uint256 claimableFundingFeeAmount);
     error UnexpectedPoolValue(int256 poolValue);
 
+    // MarketPositionImpactUtils errors
+    error InvalidImpactPoolValueForWithdrawal(uint256 poolValue);
```
xvi10: could change this to InsufficinetImpactPoolValueForWithdrawal(withdrawalAmount, poolValue)

test/guardian/testPositionImpactPool.ts
```diff
@@ -0,0 +1,131 @@
...
+import { getMarketTokenPriceWithPoolValue } from "../../utils/market";
+import { handleDeposit } from "../../utils/deposit";
+import * as keys from "../../utils/keys";
+import { usingResult } from "../../utils/use";
+import { time } from "@nomicfoundation/hardhat-network-helpers";
+import { getPositionImpactPoolWithdrawalPayload } from "../../utils/timelock";
+import { grantRole } from "../../utils/role";
+import { errorsContract } from "../../utils/error";
+
+describe("Guardian.PositionImpactPool", () => {
```
xvi10: this test should be in test/config/Timelock.ts instead

contracts/config/ConfigTimelockController.sol
Call-Sol: There is validation already for the oracle max price age. 

Oracle.sol:247L
```
uint256 maxPriceAge = dataStore.getUint(Keys.MAX_ORACLE_PRICE_AGE);
```

Oracle.sol:296L
```  
  if (validatedPrice.timestamp + maxPriceAge < Chain.currentTimestamp()) {
                revert Errors.MaxPriceAgeExceeded(validatedPrice.timestamp, Chain.currentTimestamp());
            }
```

Do we really need additional key?

xvi10: yes, because the MAX_ORACLE_PRICE_AGE may be configured to a higher value like 5 minutes

so a CONFIG_MAX_PRICE_AGE set to a smaller value would help make it safer

Call-Sol: Added config value

contracts/config/TimelockConfig.sol
```diff
@@ -332,10 +336,66 @@ contract TimelockConfig is RoleModule, BasicMulticall {
...
     function execute(address target, bytes calldata payload) external onlyTimelockAdmin {
         timelockController.execute(target, 0, payload, 0, 0);
     }
 
+    function executeWithOraclePrice(
+        address target,
+        bytes calldata payload,
+        OracleUtils.SetPricesParams calldata oracleParams
+    ) external onlyTimelockAdmin {
+        timelockController.executeAtomicWithOraclePrices(
```
xvi10: this should be timelockController.executeWithOraclePrices instead

contracts/config/TimelockConfig.sol
```diff
@@ -332,10 +336,66 @@ contract TimelockConfig is RoleModule, BasicMulticall {
...
+            target, 0, payload, oracleParams
+        );
+    }
+
+    function executeAtomicWithOraclePrice(
```
xvi10: could remove this

contracts/config/ConfigTimelockController.sol
```diff
@@ -12,13 +12,63 @@ import {OracleStore} from "../oracle/OracleStore.sol";
...
-        address[] memory executors
-    ) TimelockController(minDelay, proposers, executors, msg.sender) {}
+        address[] memory executors,
+        Oracle oracle,
+        DataStore _dataStore
+    ) TimelockController(minDelay, proposers, executors, msg.sender) OracleModule(oracle) {
+        dataStore = _dataStore;
+    }
+
+    function executeAtomicWithOraclePrices(
```
xvi10: could remove this

PR #145, opened 2025-03-20
Edge data stream verifier contract

Call-Sol:


contracts/oracle/EdgeDataStreamVerifier.sol
```diff
@@ -0,0 +1,189 @@
...
+        return keccak256(message);
+    }
+
+    /**
+     * @dev Recovers the signer's address from a signature
+     * @param messageHash The hash of the original message
+     * @param signature The signature bytes
+     * @return The address of the signer
+     */
+    function recoverSigner(
```
gdev8317: at first glance you can use ECDSA.recover instead

contracts/oracle/EdgeDataStreamVerifier.sol
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        // Recover the signer's address
+        return ecrecover(messageHash, v, r, s);
+    }
+
+    /**
+     * @dev Converts a uint256 to bytes
+     */
+    function toBytes(uint256 x) internal pure returns (bytes memory) {
```
gdev8317: maybe these helpers could be moved to utils. e.g. utils/Cast and maybe create utils/Bytes for leftPadBytes?

contracts/oracle/EdgeDataStreamVerifier.sol
Call-Sol: Sad, but I couldn't because Edge signature has v value with 27 offset, and ECDSA recover is not supporting digest with it

contracts/oracle/EdgeDataStreamVerifier.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    struct Report {
+        bytes32 feedId;
+        uint256 bid; // bid: min price, highest buy price
+        uint256 ask; // ask: max price, lowest sell price
+        uint32 timestamp;
+        int32 expo; // precision of bid&ask (negative value)
+    }
+
+    address public immutable TRUSTED_SIGNER;
```
xvi10: in other contracts, camel case is used for immutable variables, i think camel case should be used here as well

contracts/oracle/EdgeDataStreamVerifier.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+    * @dev Constructor to set the trusted signer address
+    * @param trustedSigner The address of the trusted signer
+    */
+    constructor(address trustedSigner) {
```
xvi10: could use _trustedSigner instead

contracts/oracle/EdgeDataStreamVerifier.sol
xvi10: should OpenZeppelin ECDSA.tryRecover be used instead? there are some additional validations that may make this more secure, could also check with Chaos Labs if there is anything else to note for this implementation

Call-Sol: I've added tryRecover call. But I still need this v+27 shift. Cause Edge Go code is using this offset and ECDSA tryRecover is not supporting it.

xvi10 (approved):

PR #147, opened 2025-03-30
Deployment updates

Call-Sol:
* Add sourcify support
* Support validating roles from updateRoles
* Check contract deployer
* Make summary at the script finish


scripts/updateRoles.ts
```diff
@@ -226,6 +227,25 @@ async function main() {
...
+      address: member,
+      name: contractName,
+      isCodeValidated: false,
+      signalledRoles: [role],
+      unapprovedRoles: [],
+    };
+
+    await validateSourceCode(provider, contractInfo);
+    if (!contractInfo.isCodeValidated) {
+      console.log(`❌${contractInfo.name} is not valid. Sources not match. See diff in validation folder`);
```
xvi10: should be "Sources do not match"

scripts/validateContractDeployment.ts
```diff
@@ -34,6 +31,14 @@ const roleLabels = {
 };
 
 async function main() {
+  // await validateWithSourcify({
```
xvi10: should this commented out code be removed?

scripts/validateContractDeployment.ts
```diff
@@ -49,41 +54,44 @@ async function main() {
...
-
-interface SignalRoleInfo {
-  account: string;
-  roleKey: string;
+function printResults(contractInfos: ContractInfo[]) {
+  for (const contractInfo of contractInfos) {
+    if (contractInfo.isCodeValidated) {
+      console.log(`✅${contractInfo.name} is valid`);
+    } else {
+      console.log(`❌${contractInfo.name} is not valid. Sources not match. See diff in validation folder`);
```
xvi10: should be "Sources do not match"

xvi10 (approved):

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+
+// Validate contract using sourcify verified one as reference.
+const SOURCIFY_API_ENDPOINT = "https://sourcify.dev/server/v2/contract";
+
+async function validateWithSourcify(contractInfo: ContractInfo): Promise<boolean> {
+  console.log(`Trying to validate ${contractInfo.address} via sourcify`);
+  const chainId = await hre.ethers.provider.getNetwork().then((network) => network.chainId);
+  const url = `${SOURCIFY_API_ENDPOINT}/${chainId}/${contractInfo.address}`;
+  try {
+    const path = url + "?fields=sources,compilation";
```
gdev8317: let's pass it as `params` to axios instead of manually building querystring

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+    const response = await axios.get(path);
```
gdev8317: would be great to add TS types for API responses

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+export interface DeploymentInfo {
+  contractName: string;
+  constructorArgs: string[];
+  deploymentTxHash: string;
+}
+
+export async function validateSourceCode(provider: JsonRpcProvider, contractInfo: ContractInfo) {
+  const { contractCreator } = await getContractCreationFromEtherscan(contractInfo.address);
```
gdev8317, line 72: would be good to cache positive result to avoid sending unnecessary calls. i think they can noticeably slow down updateRoles script

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+//Using streaming read cause file can be big
+const getAddressFromDeployment = async (filename: string): Promise<string | null> => {
+  if (fs.lstatSync(filename).isDirectory()) {
+    return null;
+  }
+
+  try {
+    const fileStream = fs.createReadStream(filename);
```
gdev8317: wondering would that work?

```
return JSON.parse(fs.readFileSync(filename)).address ?? null
```

or could you explain why do you need to process it this way?

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+  (address: string) =>
+  async (filename: string): Promise<boolean> => {
+    const deploymentAddress = await getAddressFromDeployment(filename);
+    return deploymentAddress === address;
+  };
+
+const findFile =
+  (searchFile: string) =>
+  (filename: string): Promise<boolean> => {
+    return Promise.resolve(filename.endsWith(searchFile));
```
gdev8317: is it necessary to return Promise<boolean> instead of just boolean?

scripts/validateDeploymentUtils.ts
```diff
@@ -0,0 +1,300 @@
...
+  };
+
+// Search recursively through all files in the `dirPath` and test it with `condition`
+// Returns filename when condition is true
+async function searchDirectory(dirPath: string, condition: (filename: string) => Promise<boolean>): Promise<string> {
```
gdev8317: at first glance this function could be sync

gdev8317: i would also suggest to use https://www.npmjs.com/package/glob

newer nodejs versions support it natively, but 20 (that we use) doesn't have it yet

scripts/validateDeploymentUtils.ts
gdev8317: i think you can also skip all files that don't end with ".json" (with this checking isDirectory wont be necessary)

gdev8317 (approved):

PR #209, opened 2025-07-17
Claim contract (from v2.2-branch)

gdev8317:


contracts/claim/ClaimHandler.sol
```diff
@@ -0,0 +1,326 @@
...
+                param.account,
+                token,
+                distributionId,
+                param.amount,
+                nextAmount
+            );
+        }
+
+        IERC20(token).safeTransferFrom(msg.sender, address(claimVault), totalTransferAmount);
+        dataStore.incrementUint(Keys.totalClaimableFundsAmountKey(token), totalTransferAmount);
```
xvi10, line 109: should totalClaimableFundsAmountKey have ID as well? e.g. totalClaimableFundsAmountKey(token, ID)

cos I think the purpose of totalClaimableFundsAmountKey is to see how much of a token is remaining to be claimed? in that case knowing how much of a token for a particular distribution ID to be claimed could be useful

contracts/claim/ClaimHandler.sol
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    // @dev deposit funds for multiple accounts and tokens in batch
+    // @param token the token to deposit
+    // @param amounts array of deposit parameters
+    function depositFunds(
+        address token,
+        uint256 distributionId,
+        DepositParam[] calldata params
+    ) external globalNonReentrant onlyConfigKeeper {
```
xvi10: this should be onlyClaimAdmin instead of onlyConfigKeeper

contracts/claim/ClaimHandler.sol
xvi10: update: i see that there is an invariant check "totalAmountLeft > IERC20(param.token).balanceOf(address(claimVault))", which may be the intended purpose of tracking this totalClaimableFundsAmountKey

i'm not sure if this invariant check is needed, but ok to leave it as is and not add ID to totalClaimableFundsAmountKey

contracts/claim/ClaimHandler.sol
gdev8317: will update for setTerms and removeTerms as well

contracts/claim/ClaimHandler.sol
gdev8317: i think we can use squid to track amount per distribution id if needed

xvi10: makes sense

PR #268, opened 2025-10-28
Update stable price

Call-Sol:
Set stable price to 12 decimals precision for tokens with 18 decimals


xvi10:
replaced by https://github.com/gmx-io/gmx-synthetics/commit/e3bcf3df8040550d0daf5cd426c6ceb0b24a2545

PR #280, opened 2025-11-18
Markets scripts

dmax10:
### Overview

Scripts to support two upcoming protocol changes:
1. Migrating `POSITION_IMPACT_EXPONENT_FACTOR` from single-parameter keys (v2.2) to dual-parameter keys with `isPositive` boolean (v2.2b)
2. Analyzing the impact of enabling `USE_OPEN_INTEREST_IN_TOKENS_FOR_BALANCE` on GM token prices

---

## Script 1: `syncPositionImpactExponentFactors.ts` - Migration Executor

**What it does**: Migrates position impact exponent factor values from the old key format to the new format across all markets.

**Why**: The v2.2b protocol upgrade changes the DataStore key structure from `(POSITION_IMPACT_EXPONENT_FACTOR, market)` to `(POSITION_IMPACT_EXPONENT_FACTOR, market, isPositive)`.

**Key behavior**: 
- Discovers all markets and reads existing values from old keys
- Writes the same value to BOTH new keys (isPositive=true and false)
- Uses Multicall3 for efficient batch reads and Config.multicall for batch writes (100 calls per tx)
- requires explicit `WRITE=true` to execute changes

---

## Script 2: `checkUseOpenInterestInTokensImpact.ts` - Impact Simulator

**What it does**: Quantifies how GM token prices would change across all markets if `USE_OPEN_INTEREST_IN_TOKENS_FOR_BALANCE` is enabled.

**How it works**:
1. Fetches baseline GM prices for all markets with current settings
2. Temporarily enables the flag (on Anvil fork) and recalculates prices
3. Outputs a comparison table showing absolute and percentage price differences per market

**Requirements**: Must run on Anvil fork (impersonates CONTROLLER to modify DataStore)

---

## Usage

**Migration**:
```bash
# Dry-run (read-only)
npx hardhat run scripts/syncPositionImpactExponentFactors.ts --network arbitrum

# Execute
WRITE=true npx hardhat run scripts/syncPositionImpactExponentFactors.ts --network arbitrum
```

**Impact Analysis**:
```bash
# Terminal 1: Start fork
anvil --fork-url $ARBITRUM_RPC_URL

# Terminal 2: Run simulation
npx hardhat run scripts/checkUseOpenInterestInTokensImpact.ts --network anvil


coltgmx (approved):
we could skip spot markets but its fine but lgtm!

PR #286, opened 2025-12-03
added new method for supporting twap orders

coltgmx:
- add new interface to support twap orders so the backward compatibility is still supported with current interfaces
- for support orders, user submit order info for individual order plus number of twaps and intervals, internally the contract will replace the original order configuring the correct valid_from time and calculate the total tokens/wnt needed. 
- frontend needs to calculate the total tokens/wnt including the execution fee and send them to order vault before submitting the request(in multicalls)
- migrated simulation function from exchange router to a new contract(SimulationRouter) due to contract size limit hit with new changes


contracts/order/OrderUtils.sol
```diff
@@ -76,6 +76,90 @@ library OrderUtils {
...
+                if (params.addresses.initialCollateralToken == wnt) {
+                    if (initialCollateralDeltaAmountTwap < executionFeeTwap) {
+                        revert Errors.InsufficientWntAmountForExecutionFee(
+                            initialCollateralDeltaAmountTwap,
+                            executionFeeTwap
+                        );
+                    }
+                    params.numbers.initialCollateralDeltaAmount = (initialCollateralDeltaAmountTwap - executionFeeTwap) / twapCount;
+                    shouldRecordSeparateExecutionFeeTransfer = false;
+                }
```
Call-Sol, line 134: Don't we need to set params.numbers.initialCollateralDeltaAmount to initialCollateralDeltaAmountTwap here? 

Otherwise it looks like user can spend vault's funds to trade

Call-Sol (changes requested):

contracts/order/OrderUtils.sol
coltgmx: > Don't we need to set params.numbers.initialCollateralDeltaAmount to initialCollateralDeltaAmountTwap here?
> 
> Otherwise it looks like user can spend vault's funds to trade

initialCollateralDeltaAmountTwap is the total amount recorded(sent by user) but depends on the number of twaps, the individual initialCollateralDeltaAmount(the collateral delta for each order) is calculated by initialCollateralDeltaAmountTwap / number of twaps. here the executionFeeTwap represents the total execution fee estimated by individual execution fee * number of twaps, we need to subtract that before the division. 
so user can never spend more than the recorded initialCollateralDeltaAmountTwap amount
the logic is similar to line 183-194 but in this case we need to do the estimation from the initialCollateralDeltaAmountTwap. overall user send one transfer to the vault to execute the twap, so the recorded initialCollateralDeltaAmountTwap is the sum of all delta amount among all the twap orders

Call-Sol: Ok, but what happens if user's collateral token is not wnt. 
User can set params.number. initialCollateralDeltaAmount to any amount whether it was supplied previously or not. 

e.g. 
* user supply 100 USDC
* user creates twap order with 10 orders amount and 100 initialCollateralDeltaAmount each.
* if orderVault contains more than 10*100 = 1000 USDC there will be 1000 USDC worth orders created having just 100 of collateral

in _createOrder there is 
```cache.initialCollateralDeltaAmount = orderVault.recordTransferIn(params.addresses.initialCollateralToken);``` line which will not be executed in case of twap order cause of checkBalance == false

coltgmx:
already merged to v2.2c

PR #302, opened 2026-01-02
SCDEV-15 Validate that chainlink price feeds are contracts

SentinelDev1:
Using the Chainlink flags contract and price feed description field to validate that chainlink price feeds are valid price feeds and contracts.


contracts/config/ConfigUtils.sol
```diff
@@ -83,6 +88,14 @@ library ConfigUtils {
             revert Errors.InvalidPriceFeed(params.priceFeed.feedAddress);
         }
 
+        if (!flags.getFlag(params.priceFeed.feedAddress)) {
```
Call-Sol, line 91: Does it implies that we can set only chainlink providers as priceFeed? 
I mean flags are only exist for chainlink providers.

SentinelDev1: Hmm, that's a good point, I didn't realize we are still using non-chainlink price feeds in some cases. If non-chainlink price feeds are also configured using initOracleConfig that would be an issue. 

I'm not sure if there is already a way to tell if a price feed is chainlink and then only check flags if the price feed uses a chainlink provider? I will investigate this further. Otherwise, the flags component could be removed.

